PR #217, opened 2015-05-28
Required to build with gcc 5.

BPaden:


cspiegel:
I've fixed the gcc5 problem by not inlining the two functions which were causing this problem, since to inline them is a violation of the C99 standard.

I'd prefer to avoid GNUisms as much as possible so as to try to be compatible with a wide range of compilers.

PR #234, opened 2016-10-09
Input handling workaround for macOS 10.12 (Sierra)

kerrymg:
### Background

Gargoyle launcher process intercepts all events and forwards relevant events to the interpreter process (mostly key down events). The interpreter validates the raw event and then translates any key down events into character codes. To do the translation, the interpreter forwards the event back to the Gargoyle process using a remote connection (`NSConnection`). The Gargoyle process then feeds the event into an `NSTextView` object using the `interpretKeyEvents:` method. Later, the interpreter process retrieves the result string out of the text view and uses it as actual input.
### Problem

Feeding raw event objects into the text view when invoked via the interpreter's remote connection doesn't work in macOS 10.12 (see `GargoyleApp` class `setWindow:charString:` which calls`GargoyleWindow` class `sendChars:` method). The exact reason is unknown at this time (security enhancement?). All events are silently ignored/discarded by the `NSTextView` object.
### Workaround

Translate the key down events into characters when they are intercepted in the window's `sendEvent:` method. To maintain original behavior only the exact events the interpreter accepts as input characters are translated. The changes should work in older Mac OS X releases.

### DIFF
--- a/garglk/launchmac.m
+++ b/garglk/launchmac.m
@@ -27,6 +27,8 @@
 #import <Cocoa/Cocoa.h>
 #import <OpenGL/gl.h>
 #import <OpenGL/glu.h>
+#include <mach-o/dyld.h>
+#import "sysmac.h"
 
 #ifdef __ppc__
 #define ByteOrderOGL GL_UNSIGNED_INT_8_8_8_8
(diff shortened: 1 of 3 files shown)


erkyrath:
This works for me. (In my homebrew-based build.) Thank you!

wmodes:
@erkyrath, do you detail the steps you took in your homebrew (Sierra I assume) build anywhere? Many of it got a trainwreck when we tried to build because of outdated libraries.

erkyrath:
See https://github.com/erkyrath/garglk/tree/buildports7 , and in particular my notes in the INSTALL file in that branch. I wound up *not* succeeding with homebrew, but I was able to install MacPorts and make it work.

https://github.com/erkyrath/garglk/releases

timofonic:
Is this solved?

kerrymg:
The work around fixes the input problem. @erkyrath has made new Gargoyle builds available, see above, but there is no official release.

sethcshort:
Using @erkyrath 's latest build (thanks!) keyboard input now works properly for me when typing in commands and using the arrow keys. However, in Floatpoint (Emily Short, http://ifdb.tads.org/viewgame?id=01efmfsk4r79mtks) at the beginning there is a requirement to "Please press SPACE to continue." Pressing the spacebar does not work here. Not sure if related to this issue.

erkyrath:
I can reproduce that. Probably related, yes...

erkyrath:
Or maybe not. I have a test game for just this sort of thing: http://eblong.com/zarf/glulx/inputeventtest.ulx

(Launch test, type GET CHAR, hit any key.)

In that test, Gargoyle *is* able to accept character input of the space bar, and all other characters. (Except tab, but I don't sweat that one.)

So Floatpoint must be in some other situation, I suppose.

Issue #213, opened 2015-03-22
Memory error when drawing inline rescaled images


erkyrath:
Running http://eblong.com/zarf/glulx/imagetest.gblorb (http://eblong.com/zarf/glulx/imagetest.inf):

Enter "image 10 right 32x64", then "image 10 right 32x100". Continue alternating these two inputs until Gargoyle crashes.

	Exception Type:        EXC_CRASH (SIGSEGV)
	Exception Codes:       0x0000000000000000, 0x0000000000000000

	Thread 0:: Dispatch queue: com.apple.main-thread
	0   libgarglk.dylib               	0x0000000100044436 gli_draw_picture + 262
	1   libgarglk.dylib               	0x000000010003dd9c win_textbuffer_redraw + 4380
	2   libgarglk.dylib               	0x000000010003a95f win_pair_redraw + 31
	3   libgarglk.dylib               	0x00000001000394fc gli_windows_redraw + 44
	4   libgarglk.dylib               	0x000000010004910e winrefresh + 14
	5   libgarglk.dylib               	0x0000000100049f05 winloop + 69
	6   libgarglk.dylib               	0x000000010004a001 gli_select + 241
	7   libgarglk.dylib               	0x0000000100026aa0 gidispatch_call + 3056
	8   git                           	0x00000001000123e3 git_perform_glk + 435
	9   git                           	0x000000010000c538 startProgram + 8808
	10  git                           	0x0000000100004e2a gitMain + 234
	11  git                           	0x000000010000507c gitWithStream + 220
	12  git                           	0x00000001000160b2 main + 98
	13  git                           	0x0000000100004d14 start + 52

erkyrath:
Sometimes there's also pixel garbage in the window:

![gargoyle-crash](https://cloud.githubusercontent.com/assets/65666/6771445/475939b0-d0b4-11e4-9b53-e08ad772faa8.png)

salty-horse:
When requested to display a scaled image, Gargoyle scales it, and stores it in a "scaled" pointer in that image slot.

When requested to draw the same image, in a different scale, Gargoyle scales it again, frees the data of the old scaled image, and stores the new one in the "scaled" pointer.

When tries to draw the first image, because it's still in the scrollback, it tries to access freed memory.

There's a "refcount" counter on the picture_t objects, but it's not used. It should be :)

salty-horse:
Here's an attempt to fix this by reference-counting images.

[...]

PR #235, opened 2016-10-17
Fix picture list clearing.

kerrymg:
Prevent possible null-pointer exceptions/dereference when clearing a picture list.

### DIFF
--- a/garglk/imgload.c
+++ b/garglk/imgload.c
@@ -74,8 +74,10 @@ void gli_piclist_clear(void)
         tmpptr = picptr;
         picptr = picptr->next;
 
-        gli_picture_decrement(tmpptr->picture);
-        gli_picture_decrement(tmpptr->scaled);
+        if (tmpptr->picture)
+            gli_picture_decrement(tmpptr->picture);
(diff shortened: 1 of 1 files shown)


erkyrath:
I expect this was a corner case neglected when https://github.com/garglk/garglk/issues/213 was fixed.

salty-horse:
Thanks! I'm curious if this is a precautionary change, or if you managed to trigger a bug.

I'm considering moving that test into `gli_picture_decrement`, as a null pointer is checked by the caller in all other instances except these two.

salty-horse:
Fixed in e3466aa3db2a917e913c67c1c5ad952b55762cad by moving the null test inside the function.

erkyrath:
Someone triggered a bug. Emily Short's _City of Secrets_ was crashing.

salty-horse:
Thanks. I will test it in case there's a flow I haven't considered.

salty-horse:
It crashed since the cover image in *City of Secrets* isn't scaled. I always tested with scaled images :(
Works fine with the fix.

PR #244, opened 2017-03-26
Improve macOS builds.

townba:
This improves macOS builds in a few ways.

- Allow building using either MacPorts or Homebrew. `gargoyle_osx.sh` will use Homebrew if available; otherwise, it will use MacPorts.
- Remove universal builds. All builds are 64-bit builds.
- Sets the version number automatically from garversion.h.
- Automatically sets internal paths correctly for all libraries and plugins used.


townba:
(I should mention that this was tested on macOS Sierra 10.12.3.)

erkyrath:
I tried a build with this patch, and I'm afraid it's not working for people with pre-10.12 Macs. See thread: http://www.intfiction.org/forum/viewtopic.php?f=38&t=21701

erkyrath:
Tried it on 10.7. The results runs on 10.7 but not on 10.12.

My conclusion is that this version of the Mac build script is suboptimal -- it would leave us having to distribute at least two Mac versions, maybe more. Sorry.

All three test builds can be downloaded at https://github.com/erkyrath/garglk/releases/tag/zarf-mac-20170328

townba:
Yeah, that wouldn't be great, especially since we know we can make builds
that work everywhere.

I wish I had access to 10.7 so I could get it to work in both cases. (I'd
be willing to try by emailing results between me and someone else, but I
doubt anyone would be jumping at that opportunity. :-) ) I'll see if I can
find what I missed in the doesn't-work-on-10.12 build.

I wonder if some of the changes could be used. In particular, large chunks
of gargoyle_osx.sh that fix up the libraries might be useful. Let me know
if you'd be willing to try some smaller incremental changes to fix things
up.

Brad

erkyrath:
Yes, I suspect it's possible to merge your homebrew/macports work with my script. I'll look at it this week if I have time.

townba:
Looks like the problem with my build script on 10.7 is with the part that identifies the libraries needed (the _echo "Copying all required dylibs..."_ section).

If you're willing, I'd be interested in what the following script outputs when run in the same directory as `gargoyle_osx.sh` for the 10.7 build that doesn't work on 10.12. I suspect the output of `otool` might be different on 10.7, which would result in the script not understanding the output, but that's really just a guess.

```bash
BUNDLE=Gargoyle.app/Contents
HOMEBREW_OR_MACPORTS_LOCATION="$(pushd "$(dirname $(which port))/.." > /dev/null ; pwd -P ; popd > /dev/null)"
echo "MacPorts location: ${HOMEBREW_OR_MACPORTS_LOCATION}"

# Get the dylibs needed.
ALL_DYLIB_PATHS="$(mktemp)"
for file in $(find "${BUNDLE}/Frameworks" "${BUNDLE}/MacOS" "${BUNDLE}/PlugIns" -type f); do
  otool -L "${file}" >> "${ALL_DYLIB_PATHS}"
done
cat "${ALL_DYLIB_PATHS}" | sort | uniq
rm "${ALL_DYLIB_PATHS}"
```

erkyrath:
Sorry about the delay... output is in https://gist.github.com/erkyrath/8ed114cd39b574215a036ad5d3c3701a .

Note that I had to change `$(mktemp)` to `$(mktemp -t dfgdfg)`, because the version of mktemp on 10.7 doesn't support the bare usage. I did this in gargoyle_osx.sh too.

erkyrath:
Okay, good news! The problem was user error! That is to say, I was stupid. I was running on a Mac with both homebrew and macports, but the homebrew packages weren't set up right. The script was trying to use homebrew anyhow.

When I set MAC_USEHOMEBREW to "no", and made some other minor changes, everything worked. See [this test build](https://github.com/erkyrath/garglk/releases/tag/zarf-mac-20170415), which is based on my https://github.com/erkyrath/garglk/tree/buildtownba-z branch.

Differences between your PR branch and mine:

- If `MAC_USEHOMEBREW` is set in the environment, the script does not attempt to set it. (This makes life easier for me doing builds in the future.)
- `-mmacosx-version-min` is set to 10.7 in Jamrules.
- The grep line to set `GARVERSION` wasn't working on 10.7 for some reason. No doubt an ancient grep bug. I changed the regexp slightly to be compatible.
- `mktemp` requires a `-t` argument on 10.7.
- I took the liberty of writing an extra section of the INSTALL file about Mac builds.

PR #250, opened 2017-04-03
Fix for non-ASCII encoded paths and file names.

kerrymg:
The macOS build now uses UTF8 encoding for all C-style strings.


erkyrath:
Works for me! (I tested it with my Mac build branch.) Thanks.

PR #251, opened 2017-04-03
improved development makefiles (linux)

gargoyle-sculptor:
I've created a new build system based on the _standard_ ``make`` system.

The patch only adds ``makefile`` and ``.gitignore`` files. No other changes.

My reason was that ``jam`` building takes such a long time (over 3.5 mins on machine for an _incremental build_ of one C source file). This is tiresome when developing.

The ``jam`` build system is fine for the _end user_ who just wants to download source, build and install, but it's not a good development build system.

Changing a single file e.g. ``garglk/window.c`` and calling top level make only builds the required objects - in this case ``libgarglk.so`` and re-links (only) all the game interpreters. It takes about 4.5 seconds on my machine.

Use like this for a top-level build. Resulting interpreters are under tads, terps/frotz etc. I have not added a ``dist`` target yet.
```
$ cd ----/garglk
$ make release # or just make (release is the default)
$ make clean debug
```
To work on a particular engine e.g. bug fixing.
```
$ cd garglk/terps/nitfol
... edit some sources
$ make release (or debug)
... only the minimum will be built - very fast, less than 0.5secs on my machine
```
### debug define
The only current difference between debug and release ``jam`` builds is the ``-g/-O2`` option to compilers.

I've added a ``-DGARGLK_DEBUG`` define which allows for test code to be kept in debug builds only. So ``make debug`` will set ``-g -DGARGLK_DEBUG`` and ``make [release]`` will set ``-O2`` (only).
### missing

- dist build (copy executables to a staging area, chmod ...)
- header file dependencies. I don't do these _yet_, but plan to. Changing a header file will not be _noticed_ by the current build system.
- multiple platform targets, testing etc. (only tested on linux, but uses standard make, which is available everywhere).


gargoyle-sculptor:
My build creates the _nitfol_ executable as ``terps/nitfol/nitfol``. This means I've trashed a shell script of the same name in that directory. I could build, say ``glknitfol``, instead, reinstate the shell file ``nitfol`` and rename ``glknitfol`` to ``nitfol`` when copying to the /dist/ area.
Alternatively I could rename the shell file to ``nitfol.sh`` and leave my build as is.

PR #254, opened 2017-04-15
MacOS build script update, try 2

erkyrath:
This is a small update of @townba's (https://github.com/garglk/garglk/pull/244). The changes allow it to work on MacOS 10.7 or later.

(Includes the TADS 3.1.3 changes from https://github.com/cspiegel/garglk/tree/203-tads-upgrade )


cspiegel:
This looks like it was branched off of my TADS upgrade branch, so will include all of its changes.

Thinking about it I honestly don't think it's a bad thing to get TADS into master right now, because at least from initial testing it looks fine, and it's just a non-starter to release without an updated TADS, so it's got to go in.

If you'd rather rebase on top of master for this PR then I'll do a separate TADS merge, otherwise, I can just merge this as-is.

erkyrath:
> Thinking about it I honestly don't think it's a bad thing to get TADS into master right now

Agreed. Go for it.

PR #258, opened 2017-04-20
Added os_file_stat implementation for MinGW.

MichelNizette:
Hoping to make myself useful here, as I seemed to understand that this was one of the blocking points for the TADS port.... I'm no C or MinGW expert at all, but I simply copied and adapted the Windows version of oss_get_file_attrs defined in tads2/osnoui.c.

With this change I successfully built gargoyle.exe and tadsr.exe on MinGW, and I tested that the correct hidden/readable/writeable file attributes are now accessible from a T3 program running in Gargoyle.

Here are the steps I followed to carry out the test (from a fresh checkout of this branch, and with the Windows version of Jam installed):

1.  I commented out the line "USESDL ?= yes ;" in the JamRules file in the base directory, so as to remove the dependency on SDL_mixer.

2.  I ran gargoyle_build.cmd. Everything built successfully except Git (I believe), which fortunately wasn't needed for this test.

3.  I ran gargoyle_win32.cmd. This copied a bunch of executables in build\dist.  The makensis.exe command in the script failed, but again whatever it's supposed to do wasn't a blocker for this test.

4.  In build\dist I copied a TADS program named "fileinfo.t3", which I created (it's attached to this message).  This program repeatedly prompts the user for file names and displays the corresponding file stats as they appear from within the TADS program.

6.  In the same folder, I ran tadsr.exe *outside* of gargoyle.exe, using the following command: "tadsr.exe -s00 fileinfo.t3".  The -s00 option was necessary to lower the file safety level so that the interpreter would allow access to the file infos I requested. Without this option, or from within Gargoyle.exe, the interpreter denied access to the stats for all files. (This was also true before my change, and may be a separate problem to look into. I don't know if it impacts only MinGW builds or all systems.)

7. From within the .T3 program, I typed various file names and checked that hidden, read-only, and read-write files are now correctly identified as such.

[fileinfo.t3.zip](https://github.com/garglk/garglk/files/945665/fileinfo.t3.zip)

### DIFF
--- a/tads/osportable.cc
+++ b/tads/osportable.cc
@@ -45,6 +45,7 @@
 #endif
 
 #ifdef GARGOYLE
+#include "osextra.h"
 #include "charmap.h"
 #include <time.h>
 #include <dirent.h>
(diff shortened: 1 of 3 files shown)


tads/osportable.cc
```diff
@@ -577,9 +577,125 @@ osfmode( const char *fname, int follow_links, unsigned long *mode,
     return ok;
 }
 
+#ifdef _WIN32
+/*
+ *   Copied and adapted from tads2/osnoui.c
+ *   Get the DOS attributes for a file
+ */
+unsigned long oss_get_file_attrs(const char *fname)
```
cspiegel: Since this function is not used outside of this file, it ought to be marked ``static``.

tads/osportable.cc
```diff
@@ -598,14 +714,13 @@ os_file_stat( const char *fname, int follow_links, os_file_stat_t *s )
     s->mode = buf.st_mode;
     s->attrs = 0;
 
+#ifdef _WIN32
+    s->attrs = oss_get_file_attrs(fname);
+#else ifndef _WIN32
```
cspiegel: This should just be ``#else`` (as written the ``ifndef _WIN32`` is extraneous and technically illegal though most compilers will ignore it).

cspiegel:
Thank you!

I just had a couple minor comments on the code, if you wouldn't mind taking a look when you have a moment.

I'd prefer to directly use the version of this function that exists in ``osnoui.c`` but unfortunately it'd require modification of that file, which I'd rather avoid so that later TADS updates can be done as cleanly as possible; and the modifications may have knock-on effects that would be a pain to deal with as well.

``osportable.cc`` comes from FrobTADS, and although it has *some* modifications, I've tried to keep it mostly in line with upstream.  However, since FrobTADS targets POSIX, MingW-specific functionality has been added as well where necessary.  But at the same time, I've got a file ``osextra.c`` which provides, at the moment, a single function which is needed, but is not part of ``osportable.cc``.  Perhaps ``oss_get_file_attrs`` would do better inside of that file, to keep ``osportable.cc`` as clean as possible.  That would require adding a new header (presumably called ``osextra.h``) which would provide the prototype for the function so that ``osportable.cc`` knows about it.  Such a header was unnecessary till now because the function provided was prototyped by TADS itself.

cspiegel:
Just one tiny point: please include ``osextra.h`` in ``osextra.c`` to ensure that a diagnostic will be issued if the prototype for ``oss_get_file_attrs`` changes.  The need to manually ensure that callers and implementations match up is one of the many weaknesses of C...  But once that's done, the change looks great.

As for the file safety level, we just stick with the TADS default which is pretty conservative.  You can use the config file to specify command-line arguments for interpreters, but that's pretty clunky.  The problem, such as it is, is that Gargoyle in a sense pretends that interpreters don't exist, so just runs each one in its default configuration, reserving config options for garglk.

For the most part this works since there's not a lot of configuration available, but that's of course not universally true, as shown by TADS here.  And there may be some other interpreter-specific tweaks that are desired (e.g. hq scalers for the graphics in Magnetic Scrolls).  

This is worth a discussion, so feel free to open a new ticket requesting the ability to change the file safety level and we can start toward seeing what kind of config tweaks we can add for interpreter-specific functionality.

MichelNizette:
Okay, I'll include the header in osextra.c - and then I'll actually retest the thing, because I did the last commit from a machine where I can't build.

Concerning the file safety setting: before opening  a new ticket as you suggest, maybe I should have a closer look at what's actually going on, because I should have thought TADS's default safety level was good enough for allowing me to do the tests I wanted to do. I thought the default was to allow reading from the current directory and/or allow read/write to the "sandbox" directory, or something like that. I think it'd be worth checking what the default exactly is, and if we can do something viable by having Gargoyle set up the sandbox directory (if it doesn't already do it), before raising a broader request for making the safety level adjustable.

MichelNizette:
And sure enough the previous version didn't compile.  :-)  It's fixed now.  Also, I think I coped for the need to define osextra.h as both a C and a C++ header file the right way (at least the build succeeded when I did it so), but let me know if this is not the right way: I'm afraid I'm already reaching territory that's unfamiliar to me here.

MichelNizette:
Concerning the file safety issue: I see in vmhostsi.cpp that the default values are

	io_safety_read_ = VM_IO_SAFETY_READWRITE_CUR;
	io_safety_write_ = VM_IO_SAFETY_READWRITE_CUR;

and aren't modified anywhere in the entire code, except optionally via command-line parameters passed to tadsr.exe, so those default values should be the ones prevailing in Gargoyle.

My interpretation of those values, reading both the comments in vmhost.h and the TADS 3 System Manual chapter on file safety, is that the default settings should allow read/write access to the sandbox directory, which if left unspecified should default to the directory of the .t3 executable.

Using my fileinfo.t3 test program with maximum permissions, I confirm that "." as seen from within the TADS program represents the directory of the fileinfo.t3 executable.  But then, running the program with default safety levels, I'm not allowed to access any file from that directory, in contradiction with the documentation.

Setting explicitly the sandbox or default directory via the -sd or -d option doesn't solve it.

The resolution of the issue, therefore, may not be so much a matter of allowing users to lower the safety level as a matter of correctly applying the default safety level in the first place.  What do you think?

cspiegel:
For the ``extern "C"`` bit it's usually just guarding the whole file (see ``tads/tads2/os.h`` for an example).  It's not wrong the way it is now, but if future code is added that will be used from C++, it'd be simpler to just wrap the whole thing in:

```c++
#ifdef __cplusplus
extern "C" {
#endif

[header code]

#ifdef __cplusplus
}
#endif
```
Regarding the permissions issue, does entering (say) ``fileinfo.t3`` when running your program fail?  As in, it's not allowed to read it?  I don't have a Windows build, but in Linux, TADS is able to stat the file fine, without setting any safety level.  If that's the case, I'll try to do a MinGW build and test there, too.

MichelNizette:
Yes, entering fileinfo.t3 fails from denial of access permission, as well as any other file in the same directory. Interestingly, entering a nonexistent file name fails too, though when the safety level is lowered nonexistent files are correctly reported as such. 

I could have a look at what TADS is actually doing when enforcing those safety checks, and how it translates to OS-specific calls. Unfortunately, I'm afraid I may not have a change to do this before a few days.

MichelNizette:
Having had a closer look at the file safety issue and on what OS-specific functions it depends, I'm under the impression that many of the definitions in tads/tads2/osnoui.c would benefit from being compiled with -DMSDOS, under MinGW.  I'll change the Jam file to achieve that and try to rebuild and see if things improve, when I get a chance to.

MichelNizette:
Unfortunately, adding -DMSDOS (and -DT_WIN32) didn't solve the safety issue. I'd need to dig deeper.

cspiegel:
For the time being then, if you're comfortable with it, I can merge this.  It's at least better than what currently exists!

MichelNizette:
Yes, I think this can be merged (I just added back an #ifdef I had inadvertently deleted).  The safety issue is a different one anyway, which pre-dated this change.  I'll keep looking into it, but again I may not have a chance to do it before a few days, and there's no reason to delay things further with the current change.

PR #264, opened 2017-05-07
Try building in GCC 4.8 and 6

curiousdannii:
I'm not sure if this will work, lets see what Travis does...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,21 @@
-script: jam -sC++=g++-6 -sCC=gcc-6 -j4
 addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
     packages:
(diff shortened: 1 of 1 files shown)


curiousdannii:
I guess the build matrix works because the new scottfree failed only in 4.8!

cspiegel:
The timing really couldn't be better... I just pushed that Scott change half an hour ago.  I'll update the Jamfile to specify C99 mode for it, which should get it working with 4.8.

cspiegel:
OK, C99 has been enabled.  With luck merging master should cause the next build to succeed.

cspiegel:
With the scott C99 fix, both 4.8 and 6 now pass. Thanks for doing this!

PR #266, opened 2017-05-09
[macOS] Add text-to-speech (TTS) support.

kerrymg:
Initial macOS Text-to-Speech support using Cocoa native API.

Don't forget to enable the TTS option (USETTS) when building.

### DIFF
--- a/garglk/ttsmac.m
+++ b/garglk/ttsmac.m
@@ -29,17 +29,56 @@
 
 #import "sysmac.h"
 
+// a queue of phrases to feed to the speech synthesizer
+static NSMutableArray<NSString *> * phraseQueue = nil;
+static NSRange purgeRange;
+
+@interface SpeechDelegate : NSObject <NSSpeechSynthesizerDelegate>
+{
+}
+- (void) speechSynthesizer: (NSSpeechSynthesizer *) sender
+         didFinishSpeaking: (BOOL) finishSpeaking;
+@end
+
+@implementation SpeechDelegate
+{
+}
+- (void) speechSynthesizer: (NSSpeechSynthesizer *) sender
+         didFinishSpeaking: (BOOL) finishSpeaking
+{
+    if (finishSpeaking == NO)
+    {
+        // speaking stopped abruptly (see gli_tts_purge())
+        [phraseQueue removeObjectsInRange: purgeRange];
+    }
+    else
+    {
+        // remove the head of the list which just completed speaking
+        if (phraseQueue.count > 0)
+        {
+            [phraseQueue removeObjectAtIndex: 0];
+        }
+    }
+
+    // speak the next phrase
+    if (phraseQueue.count > 0)
+    {
+        [sender startSpeakingString: phraseQueue[0]];
+    }
+}
+@end
+
 #define TXTSIZE 4096
 static glui32 txtbuf[TXTSIZE + 1];
 static size_t txtlen;
 
-static NSMutableString * ttstext = nil;
 static NSSpeechSynthesizer * synth = nil;
 
 void gli_initialize_tts(void)
 {
     if (gli_conf_speak)
     {
+        // spawn speech synthesizer using the default voice
         synth = [NSSpeechSynthesizer new];
         if (!synth)
         {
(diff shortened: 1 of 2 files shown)


garglk/ttsmac.m
```diff
@@ -0,0 +1,155 @@
...
+        {
+            [ttstext appendString: text];
+            [text release];
+        }
+
+        txtlen = 0;
+    }
+}
+
+void gli_tts_speak(const glui32 *buf, size_t len)
```
cspiegel: If I'm reading this function right, it will only ever extend ``ttstext``, but not actually speak anything.  Speaking will only occur on ``gli_tts_flush()``.  However, that function is only called when there's user input.  ``gli_tts_speak()`` is called for each character written to a text buffer window, so the speech-dispatcher and Windows backends flush whenever an end of sentence marker is reached (thus the check for period, newline, etc).  This ensures that text is spoken (at a hopefully appropriate point) regardless of whether user input is being requested.

kerrymg: You are correct that nothing is spoken until `gli_tts_flush()` is called. Calling `[NSSpeechSynthesizer startSpeakingString:]` doesn't queue the text but immediately cancels whatever is being spoken and starts speaking the new message; calling it on a character by character basis wouldn't work. So, I opted to buffer everything until the next input prompt and then output all the text in a single call.

The tests I ran with a few games seems that it works relatively OK in practice (I wouldn't have made the PR otherwise). The macOS speech synthesizer is aware of paragraph breaks and pauses appropriately. This is one of the reasons I asked for feedback especially if someone could compare how it behaves against Linux or Windows builds.

cspiegel: > I think where it breaks down is handling of things like the - More - prompt.

The original implementation was done a long time ago (probably by Tor Andersson), so I can only guess as to the intention behind it, but I suspect it has to do with this.  That is, it's certainly possible for a Glk program to write out text but then do something else for a bit before asking for input.  In addition, I don't think the current implementation flushes on single-character input, only for line input, which would mean a “press any key to continue” prompt might not trigger a flush.

Some of this can be alleviated with well-placed calls to ``gli_tts_flush()``, e.g. at character input or when the “more” prompt is displayed.  Then, the only likely issue for the Mac implementation is text displayed followed by a pause, which would result in a delay before the text is spoken.

Realistically, the vast majority of games will request input immediately after displaying text, so this is mainly a theoretical problem, and if the Mac speech synthesis simply won't buffer input, then the manual buffering here is the only option.  But I think adding more calls to ``gli_tts_flush()`` will at least improve the situation, and won't negatively effect Linux or Windows.

kerrymg: Hmm... A quick look at the `- More -` prompt handling shows it's somewhat involved. Mainly because it can be dismissed by mouse clicks, manually scrolling the buffer window, or character input. I'll have to study input and text output handling some more.

I think a better solution on Mac (that would more closely mimic Linux/Windows TTS behavior) would be to use a queue/FIFO and have a separate thread feed the speech synthesizer while draining the queue. The synthesizer notifies a delegate of events, in this case when it stops speaking, so we can feed it the next "sentence" fragment. Calls to `gli_tts_speak()` would fill the queue and signal the speech thread if not already speaking.

This is obviously a more involved implementation but it may not be as bad as it sounds.

cspiegel: Grid windows are not spoken, that's correct.

The current TTS implementation evolved from the initial, unused TTS, which means that there's probably room for improvement; so if there's anything you run across that feels “off” in how it's done, it's worth discussing whether the approach should be changed.

kerrymg:
I managed to get the Linux build up and running (Ubuntu 17.04 VM) and compared that build's TTS support against my latest TTS implementation for macOS. They both behave pretty much identically over the sample game files I used for testing (doesn't mean there isn't room for improvement).

cspiegel:
OK, I'll merge this in.  As it stands I suspect all of the TTS code has some room for improvement.  At some point I'd like to pull out some of the duplicated (or near duplicated) code across all the implementations, but for the time being, I'll be happy to have something that works on all the major platforms.

Thank you for working on this!

PR #270, opened 2017-06-03
Fullscreen mode for Windows

flabbergeist:
Enabled by adding "fullscreen 1" to the configuration file (garglk.ini).

### DIFF
--- a/garglk/syswin.c
+++ b/garglk/syswin.c
@@ -315,18 +315,32 @@ void winopen()
 {
     HMENU menu;
 
-    int sizew = gli_wmarginx * 2 + gli_cellw * gli_cols;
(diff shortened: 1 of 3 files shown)


garglk/syswin.c
```diff
@@ -315,18 +315,32 @@ void winopen()
 {
     HMENU menu;
 
-    int sizew = gli_wmarginx * 2 + gli_cellw * gli_cols;
-    int sizeh = gli_wmarginy * 2 + gli_cellh * gli_rows;
+    int sizew = 0;
+    int sizeh = 0;
+	DWORD dwStyle = 0;
```
cspiegel: Spaces should be used instead of tabs for indentation.

Issue #272, opened 2017-07-18
Feature Request: Ctrl+LeftArrow / Ctrl+RightArrow


shack19:
Pretty much the only issue I'm really affected by is that Ctrl+LeftArrow and Ctrl+RightArrow don't work to move the cursor between words. Is this something that has been discussed in the past?

cspiegel:
Merged in #291, so this will be in the next release.

PR #291, opened 2017-11-03
Support control-left and control-right for skipping words.

cspiegel:
This addresses issue #272, which requests control-left and control-right to skip words.  There are a couple things:

* Pseudo keycodes are used to represent control left and control right. A more “proper” approach would probably be to fully support modifiers, but that would cause a somewhat large change in the input system that I'm not ready to do.
* EFL, Windows, and gtk all have support here, but OS X does not. I don't have access to an OS X machine, nor do I know how it's supposed to behave, whether this ticket applies to it at all.  If anybody wants to contribue OS X support that would be great but I will also merge without it if not.
* Only spaces count as word separators.  I don't think this is really a problem.

### DIFF
--- a/garglk/glk.h
+++ b/garglk/glk.h
@@ -482,5 +482,7 @@ extern void garglk_set_reversevideo_stream(strid_t str, glui32 reverse);
 #define keycode_Erase               (0xffffef7f)
 #define keycode_MouseWheelUp        (0xffffeffe)
 #define keycode_MouseWheelDown      (0xffffefff)
+#define keycode_SkipWordLeft        (0xfffff000)
+#define keycode_SkipWordRight       (0xfffff001)
 
 #endif /* GLK_H */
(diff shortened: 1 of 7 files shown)


kerrymg:
If nobody minds I'll add support for this in macOS (OS X).

However, should the key combination remain `control-left/right`? The `option-left/right` key combination is used system wide to skip words.

cspiegel:
Thanks for taking this on. 

If option is the standard on OS X, then I suggest doing that. I think it's more important for Gargoyle to fit in on a platform than be consistent across platforms.

curiousdannii:
Control on Windows/Linux equals Option in OSX. Just one of those dumb things.

At least most keyboards I've seen have stopped putting the Fn in the very bottom left where Control should be...

kerrymg:
@curiousdannii OS X is inconsistent, some Windows/Linux control keys map to command others to alt/option, while some others use control (like control-C in a Terminal).

I implemented the feature using alt/option to skip words. The control-left/right combinations are used for Mission Control (nee Exposé) keyboard shortcuts.

@cspiegel Should I do a PR off the main repo or your branch?

cspiegel:
@kerrymg Go ahead and do it off my branch.

Issue #299, opened 2017-11-11
CJK fonts mostly don't work


alysdexia:
Noto Serif CJK SC, Noto Serif CJK TC, Noto Sans CJK SC, Noto Sans CJK TC, Noto Sans Mono CJK SC, Noto Sans Mono CJK TC, MingLiU-ExtB, and Microsoft JhengHei UI fail to the default.  DengXian load the bold and bold oblique sans-serifs but regular and italic fail to the default serif.

cspiegel:
For one of the affected fonts, can you please link to exactly which font files you installed to provide all weights for that font (i.e. Roman/Bold/Italic/Bold Italic, or whatever the appropriate names are for CJK fonts), as well as to a game file that exhibits the problem?

alysdexia:
Pretty well-known: https://www.google.com/get/noto/.  The other fonts belong to Windows.

It looks like the very few games written in CJK use none of the formats garglk supports, but usually Hypercard-like engines like Ren'py (https://www.renpy.org/), Twine (http://twinery.org/), and INterpreter of Simple TExt ADventure (https://instead.syscall.ru/index.html) or HTML engines like Squiffy (http://textadventures.co.uk/squiffy).  Anyhow I still expect a CJK font to display its Latin code for European languages.  Dreamhold (http://ifdb.tads.org/viewgame?id=3myqnrs64nbtwdaz) was the game I tried in.

kerrymg:
I believe I figured out what the problem is. Unfortunately, I don't have a simple solution.

The main issue is that these CJK fonts are TrueType/OpenType font collections (`.ttc` extension with multiple fonts within a single file) and Windows treats them differently in the Registry. Gargoyle uses the fonts' Registry entries to locate the actual font file to pass to Freetype. It does this by locating a Registry key of the form "`$(FONTNAME)  (TrueType)`". This matching works for simple `.ttf` fonts but the `.ttc` fonts use keys named "`$(FONT1) & $(FONT2) & ... (TrueType)`" (each font within the collection file separated by ` & `). Because of the new naming convention the search for the actual font file never succeeds. (This is the case under Windows 10).

Refer to `garglk/winfont.c` source file for full details.

I wonder if we could change the font options in `garglk.ini` to pass the actual font filename. Admittedly an ugly and not very user friendly change.

cspiegel:
Fixed per #303.

PR #303, opened 2017-12-24
Update Windows font handling and support OpenType fonts

kerrymg:
Made changes to the Windows build so it can properly locate OpenType fonts during font enumeration and loading. Tested font loading with the OpenType version of Source Code Pro and the Noto Sans/Serif CJK CS fonts (on a Windows XP VM - presumably later versions of Windows will work also).

Should fix issue #299 

I also took the liberty to upgrade the Freetype library to the latest version along with some of Freetype's dependencies (zlib and libpng).


garglk/fontwin.c
```diff
@@ -228,6 +191,62 @@ int CALLBACK propfont(
     return 1;
 }
 
+static void make_font_filepath(const char *filename, char *filepath, size_t length)
+{
+    // create the absolute path to the font file
+    filepath[0] = '\0';
+    if (!strstr(filename, ":") && getenv("SYSTEMROOT"))
```
cspiegel: As long as this is being touched, it really ought to be updated to be safe against buffer overflows.  There are lots of places in Gargoyle that are (potentially) susceptible to overflows, and they should all be addressed at some point, but at least we can plug some of them now.  I'd recommend something like:

```c
if (!strstr(filename, ":") && getenv("SYSTEMROOT"))
    snprintf(filepath, length, "%s\\Fonts\\%s", getenv("SYSTEMROOT"), filename);
else
    snprintf(filepath, length, "%s", filename);
```

garglk/fontwin.c
```diff
@@ -228,6 +191,62 @@ int CALLBACK propfont(
...
+    char face[256];
+    char filename[256];
+
+    // open the Fonts registry key
+    if (RegOpenKeyEx(HKEY_LOCAL_MACHINE, "Software\\Microsoft\\Windows NT\\CurrentVersion\\Fonts", 0, KEY_QUERY_VALUE, &hkey) != ERROR_SUCCESS)
+        return FALSE;
+
+    // check for a TrueType font
+    strncpy(face, facename, sizeof face);
+    strcat(face, " (TrueType)");
```
cspiegel: As above, let's take the opportunity to make this safe:

```c
snprintf(face, sizeof face, "%s (TrueType)", facename);
```

garglk/fontwin.c
```diff
@@ -228,6 +191,62 @@ int CALLBACK propfont(
...
+    if (RegQueryValueEx(hkey, face, NULL, NULL, (PBYTE)filename, &size) == ERROR_SUCCESS)
+    {
+        make_font_filepath(filename, filepath, length);
+        RegCloseKey(hkey);
+        return TRUE;
+    }
+
+    // check for an OpenType font
+    strncpy(face, facename, sizeof face);
+    strcat(face, " (OpenType)");
```
cspiegel: Ditto `snprintf`.

kerrymg:
Here is a screenshot of Gargoyle displaying Chinese characters:

screen shot 2017-12-24 at 12 42 55

(The font is Noto Sans CJK SC)

cspiegel:
Awesome, thank you for tackling this.

Issue #304, opened 2017-12-26
Expose the custom Glk functions


curiousdannii:
I have [started specifying Gargoyle's Glk extensions](http://curiousdannii.github.io/if/gargoyle.html) so that they can be exposed directly to the VMs, and so that authors can use them. It would be great if Gargoyle could include them in the next release.

Steps:

- [ ] Add the Gestalt code
- [ ] Expose the functions in the gi_dispa code
- [ ] (Not essential) Convert terps which currently tests `#ifdef GARGLK` to instead test the Gestalt code. This would be of greatest value to terps which include Gargoyle support upstream

I will post a unit test file soon.

Edit: tests available:  [gargoyle-text.inf](https://github.com/curiousdannii/if/blob/master/tests/gargoyle-text.inf), [gargoyle-text.ulx](https://github.com/curiousdannii/if/blob/master/tests/gargoyle-text.ulx)

PR #305, opened 2017-12-26
Expose the Garglk text formatting functions

curiousdannii:
Expose the custom Glk functions (#304)


curiousdannii:
Screenshot of the test file working

It works! Gaps between some of the colours aren't ideal, but they're not caused by this patch.

Issue #315, opened 2018-04-19
Mouse Input Events broken by retina/zoom change?


angstsmurf:
I haven't tested this properly, but it seems that at least clicking on the compass rose in Counterfeit Monkey no longer works after b4a966f on my retina Mac.

hkleinsorgen:
Observed on Linux and Windows, too. I've fixed this issue, see [pull request  318](https://github.com/garglk/garglk/pull/318)

angstsmurf:
Thanks a lot for fixing this!

PR #318, opened 2018-04-22
Fix issue 315: Mouse Input Events broken by retina/zoom change

hkleinsorgen:
This fixes issue 315 (Mouse Input Events broken by retina/zoom change) and a related issue (graphics were not correctly resized when using a zoom factor > 1)

I also replaced all occurrences of
  ` someExpression * gli_zoom + 0.5`
and 
 ` someExpression / gli_zoom + 0.5 `
with two macros: 
  `gli_zoom_int(someExpression)`
  `gli_unzoom_int(someExpression)`

### DIFF
--- a/garglk/config.c
+++ b/garglk/config.c
@@ -623,21 +623,21 @@ void gli_startup(int argc, char *argv[])
 
     if (gli_hires)
         gli_zoom *= gli_backingscalefactor;
-    gli_baseline = gli_baseline * gli_zoom + 0.5;
+    gli_baseline = gli_zoom_int(gli_baseline);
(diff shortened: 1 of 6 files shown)


cspiegel:
I couldn't reproduce the issue in master, so I can't verify this fixes it for me, but at least the issue is not present in this branch either.  I'll merge it.

@angstsmurf, if you have a chance, can you verify that this did fix your issue?  If not, I'm fine reverting these two patches for the time being and getting them updated after the release (which really will happen, I promise!)

angstsmurf:
Yes, this seems to fix it for me. Perhaps #315 only manifests on HiDPI screens?

PR #319, opened 2018-05-05
Expand game filename to full path before startup

muphil:
This allows using relative paths, like gargoyle if/foo.blb,
before this patch, gargolye changes working directory with glkunix_set_base_file
without modifying the game file path, thus it can't find the game file.

### DIFF
--- a/garglk/main.c
+++ b/garglk/main.c
@@ -24,6 +24,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <math.h>
+#include <limits.h>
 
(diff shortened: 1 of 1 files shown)


cspiegel:
For some reason MinGW doesn't include realpath(), so this fails to build there.  It looks like you ought to be able to just use `_fullpath` instead (https://msdn.microsoft.com/en-us/library/506720ff.aspx) if `WIN32` is defined.

In addition, please place opening braces on their own lines (Allman style) to match other files in the project.

PR #328, opened 2018-08-30
Switch bundled fonts to Noto Serif and Go Mono.

cspiegel:
These fonts have much wider Unicode support than Charter and Luxi.

Windows test build: [gargoyle-2018.1-windows.zip](https://github.com/garglk/garglk/files/2334480/gargoyle-2018.1-windows.zip)


nickbe:
Well to me this font doesn't look even remotely as nice as before.
But then is it possible to use my own fonts somehow?

cspiegel:
> Well to me this font doesn't look even remotely as nice as before.
> But then is it possible to use my own fonts somehow?

Yes, you've always been able to use your own fonts.  In your `garglk.ini`/`garglkrc` file, you can use the `propfont` and `monofont` keys to set proportional and monospaced fonts.

On Unix systems you can use `$HOME/.garglkrc`, and on Windows, find the location of `gargoyle.exe `(probably something like `C:\Program Files\Gargoyle`), and there should already be a `garglk.ini` file there next to it.  This is just a text file that you can edit with any text editor.  If you don't have the file, it's available in the repository here as `garglk/garglk.ini`.

Configuring Gargoyle isn't exactly user friendly at the moment.  I've been toying around with some ways to improve it recently, but I'm not sure how long it'll be before anything comes about.

nickbe:
Where does gargoyle expect the fonts, and how exactly are they declared in the config?

cspiegel:
It uses whatever fonts your system provides.  Just name them as you would expect, e.g.:

    propfont Sabon
    monofont Liberation Mono

nickbe:
Thanks!

wyatt8740:
This also massively upped the size of a statically linked interpreter binary. (around 750K for a statically linked 'terp became around 2.8M – `fontdata.o` became 2.1M of that).
If anyone else is here trying to figure out why their binary got so much bigger, note that you can disable the baking-in of fonts entirely with `BUNDLEFONTS=no` in `Jamrules`.

Issue #204, opened 2015-03-22
Sound notification events are not delivered if the player does nothing


GoogleCodeExporter:
```
The attached Inform 7 source (and compiled Glulx Blorb) is a minimal example 
that, when the command "chime" is entered, plays a sound. A notification event 
is requested, and when this arrives, the game cancels line input, prints a 
message and then re-starts line input.

However, when run under Gargoyle, the sound notification event is not received 
by the game until the first time after the sound has finished that the user 
generates some user interface activity (e.g. by pressing a key). If the user 
does nothing, the event is never received. From my reading of the Glk 
specification, this is not what was intended.

(See also this thread: http://www.intfiction.org/forum/viewtopic.php?f=7&t=2848)
```

Original issue reported on code.google.com by `dav...@davidkinder.co.uk` on 16 Feb 2013 at 9:58

Attachments:
- [Sound Notifications.gblorb](https://storage.googleapis.com/google-code-attachments/garglk/issue-204/comment-0/Sound Notifications.gblorb)
- [source.txt](https://storage.googleapis.com/google-code-attachments/garglk/issue-204/comment-0/source.txt)

PR #336, opened 2018-10-21
Updated SDL Sound code

angstsmurf:
This implements the extended GLK sound functions:

glk_schannel_create_ext()
glk_schannel_play_multi()
glk_schannel_set_volume_ext()
glk_schannel_pause()
glk_schannel_unpause()

What is still missing is working sound and volume notifications. They won't fire until some other event, such as the player pressing a key or moving the mouse, comes along, as reported in #204. This is implemented in #337, if only for Windows and macOS at present.

These changes only touch the SDL sound backend, and I have not been able to test whether FMOD sound still works.

I've made a test program [here](https://github.com/angstsmurf/soundtest/releases).

### DIFF
--- a/garglk/garglk.h
+++ b/garglk/garglk.h
@@ -30,6 +30,10 @@
 
 #include <stddef.h>
 
+#ifdef GARGLK_USESDL
+#include <SDL_timer.h>
+#endif
+
 #include "gi_dispa.h"
 
 /* First, we define our own TRUE and FALSE and NULL, because ANSI
@@ -564,6 +568,18 @@ struct glk_schannel_struct
     glui32 loop;
     int notify;
     int buffered;
+
+#ifdef GARGLK_USESDL
+    int paused;
+
+    /* for volume fades */
+    int volume_notify;
+    int volume_timeout;
+    int target_volume;
+    double float_volume;
+    double volume_delta;
+    SDL_TimerID timer;
+#endif
 
     gidispatch_rock_t disprock;
     channel_t *chain_next, *chain_prev;
--- a/garglk/sndsdl.c
+++ b/garglk/sndsdl.c
@@ -48,6 +48,9 @@
 #define giblorb_ID_WAVE (giblorb_make_id('W', 'A', 'V', 'E'))
 
 #define SDL_CHANNELS 64
+#define GLK_MAXVOLUME 0x10000
+#define FADE_GRANULARITY 100
+
 
 static channel_t *gli_channellist = NULL;
 static channel_t *sound_channels[SDL_CHANNELS];
@@ -61,7 +64,7 @@ void gli_initialize_sound(void)
 {
     if (gli_conf_sound == 1)
     {
-        if (SDL_Init(SDL_INIT_AUDIO) == -1)
+        if (SDL_Init(SDL_INIT_AUDIO|SDL_INIT_TIMER) == -1)
         {
             gli_strict_warning("SDL init failed\n");
             gli_strict_warning(SDL_GetError());
@@ -90,10 +93,16 @@ void gli_initialize_sound(void)
         }
         int channels = Mix_AllocateChannels(SDL_CHANNELS);
         Mix_GroupChannels(0, channels - 1 , FREE);
+        Mix_ChannelFinished(NULL);
     }
 }
 
 schanid_t glk_schannel_create(glui32 rock)
+{
+    return  glk_schannel_create_ext(rock, GLK_MAXVOLUME);
+}
+
+schanid_t glk_schannel_create_ext(glui32 rock, glui32 volume)
 {
     channel_t *chan;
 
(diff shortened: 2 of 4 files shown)


garglk/sndsdl.c
```diff
@@ -516,7 +647,21 @@ static glui32 play_mod(schanid_t chan, long len)
     FILE *file;
     char *tn;
     char *tempdir;
-    int music_busy;
+    int music_busy, loop;
+
+    if (chan == NULL)
+        gli_strict_warning("MOD player called with an invalid channel!");
```
cspiegel: Since `chan` is dereferenced below, I presume the function should return here instead of continuing.

garglk/sndsdl.c
```diff
@@ -247,54 +263,163 @@ glui32 glk_schannel_play(schanid_t chan, glui32 snd)
...
+
+    chan->float_volume = (double)chan->volume;
+    chan->volume_delta = (double)(chan->target_volume - chan->volume) / FADE_GRANULARITY;
+
+    chan->volume_timeout = FADE_GRANULARITY;
+
+    if (chan->timer)
+        SDL_RemoveTimer((SDL_TimerID)chan->timer);
+
+	chan->timer = (int)SDL_AddTimer((Uint32)(duration / FADE_GRANULARITY), volume_timer_callback, (void *)chan);
```
cspiegel: This is indented with a tab instead of 4 spaces.

cspiegel: In addition, `chan->timer` is an `int` and `SDL_AddTimer` returns a pointer.  On many 64-bit systems this could result in the truncation of the pointer.  `timer` should probably just be made an `SDL_TimerID` instead of `int`.  I think you'd have to add a macro, e.g. `-DGARGLK_USESDL` when building with SDL, so you can conditionally include `SDL/SDL_timer.h` in `garglk/garglk.h` and conditionally define `timer`.  In fact, you could probably conditionally include most/all of `struct glk_schannel_struct` so the members don't exist unless SDL is being used for sound.

garglk/sndsdl.c
```diff
@@ -303,17 +428,19 @@ static void music_completion_callback()
 static void sound_completion_callback(int chan)
 {
     channel_t *sound_channel = sound_channels[chan];
-    if (!sound_channel || Mix_Playing(chan))
-    {
-        gli_strict_warning("sound callback failed");
+
+        if (!sound_channel)
```
cspiegel: This is indented too far.

garglk/sndsdl.c
```diff
@@ -303,17 +428,19 @@ static void music_completion_callback()
...
+        {
+        fprintf(stderr, "sound_completion_callback called with invalid sound_channel %d\n", chan);
```
cspiegel: This ought to be done with `gli_strict_warning`, building up a message first with `snprintf`.  Ideally `gli_strict_warning` would take a format string and variable number of arguments.  I've added #355 to do this.

garglk/sndsdl.c
```diff
@@ -582,36 +732,75 @@ glui32 glk_schannel_play_ext(schanid_t chan, glui32 snd, glui32 repeats, glui32
...
         default:
             gli_strict_warning("schannel_play_ext: unknown resource type.");
     }
 
-    return 0;
+    /* if channel was paused it should be paused again */
+    if (result && paused)
+    {
+        if (paused)
+            fprintf(stderr, "glk_schannel_play_ext: pausing channel again\n");
```
cspiegel: Is this message necessary?  If so, presumably it should be done via `gli_strict_warning`.

garglk/sndsdl.c
angstsmurf: Thanks for the review! I actually think SDL_TIMERID changes from a pointer to an int between SDL versions, which is why the code ended up this way, but not really an excuse. I'll see if I can work out how to fix it properly.

angstsmurf:
Right, I think I've done all the requested changes, and it seems to build fine on my Mac at least. Someone should test if it still works with FMOD sound.

angstsmurf:
A weird thing is that if I try make the other SDL-specific members of glk_schannel_struct ([sdl_rwops, sdl_memory and sdl_channel](https://github.com/garglk/garglk/blob/check-babel-properly/garglk/garglk.h#L554-L556)) also conditional SDL sound, most games will break and only show a blank screen.

angstsmurf:
Squashed all commits into one. Hopefully this makes the changes more readable.

garglk/garglk.h
```diff
@@ -32,6 +32,10 @@
 
 #include "gi_dispa.h"
 
+#ifdef GARGLK_USESDL
+#include "SDL/SDL_timer.h"
```
cspiegel: One last thing here: I just tried a Windows build, and it fails because the include path points directly to the SDL directory, meaning that `SDL_timer.h` has to be included directly, not using the `SDL/` prefix.  This seems to be the proper way to do it on Unix systems too, as `pkg-config sdl --cflags` spits out `-I/usr/include/SDL`.

I'd also request using angle brackets instead of quotes since SDL is nominally a system library, even though on Windows it's being included directly, e.g.:

```c
#include <SDL_timer.h>
```
The test program looked good, so with this one change, I'll merge it in.

angstsmurf: Fixed and squashed again.

angstsmurf:
Rewrote the PR description and commit message.

cspiegel:
Thanks for implementing this!

PR #342, opened 2019-01-23
Replace /bin/ash shell with /bin/sh

jumapico:
Use /bin/sh as the majority of the scripts does.
Currently the ash shell is not widely used and was superseded by dash
shell in many linux distributions.


cspiegel:
This script isn't used by Gargoyle.  It's just part of the upstream nitfol package, and I'd rather not modify it since we're not making use of it.  This should really be sent upstream, although I'm not sure Nitfol is under active development anymore.

PR #344, opened 2019-05-20
Actually don't require Babel if USEBABEL='no'

wyatt8740:
An error would pop up at link-time if `USEBABEL` was set to `no` in `Jamrules`.
This fixes it.

### DIFF
--- a/garglk/config.c
+++ b/garglk/config.c
@@ -617,5 +617,7 @@ void gli_startup(int argc, char *argv[])
     gli_initialize_sound();
 
     winopen();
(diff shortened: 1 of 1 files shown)


cspiegel:
I dug into this a bit and found that the root cause was the Jamfiles not properly checking the `USEBABEL` variable.  I've made an alternative fix in #350.  

This way the source doesn't need to be changed, since it was written in such a way that it just stubs in an empty Babel initialize function if Babel support is turned off.

Thanks for finding this!

PR #352, opened 2019-11-03
Mac build script tweak

erkyrath:
Build correctly on MacOS 10.15 *or* if you have the latest Xcode installed on an earlier OS.

SDK_VERSION is 10.15 in these cases. The case statement was only checking for 10.14. Now it checks for 10.14 to 10.19, which is a temporary fix in some sense, but I'm sure we'll be revising the script before MacOS 10.20 ships.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -132,4 +132,81 @@ problem. That library is trying to load out of MacPorts or Homebrew, and
 the interpreter will fail on most other Macs (which do not have those repos
 installed).
 
+- MacOS code signing and notarization
 
+For a fully accessible build on MacOS 10.15 (Catalina), you need to both
+sign and notarize the app before packaging it up. (Unsigned builds still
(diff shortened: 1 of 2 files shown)


kerrymg:
I was going to issue a similar PR but I see I don't have to now. However, may I suggest a further tweak to the build script (I don't think it warrants a separate PR)?

Change the final disk image creation command (line 126) from:

`hdiutil create -ov -srcfolder Gargoyle.app/ "gargoyle-$GARVERSION-mac.dmg"`

to

`hdiutil create -fs "HFS+J" -ov -srcfolder Gargoyle.app/ "gargoyle-$GARVERSION-mac.dmg"`

The addition of the `-fs "HFS+J"` option forces creation of an HFS+ disk image instead of an APFS image (the default in newer macOS versions). Allows the image to be mounted on systems without APFS support.

erkyrath:
Good point, thanks.

PR #353, opened 2019-11-04
Update Alan3 to Build2023.

cspiegel:


cspiegel:
The official current release (according to https://www.alanif.se/) is 3.0beta6, so I've created a branch based off of that instead.

PR #357, opened 2019-11-16
Fix font file lookup on Windows

### LINKED ISSUE #351: Cannot set fonts
Using the latest 2019.1 release on Windows 10.

I put the following in garglk.ini:

```
monofont      Fantasque Sans Mono
monosize      16

propfont      Arial
propsize      16
```

Here's what I got:

2019-11-03 22_10_00-Laid Off from the Synesthesia Factory - Katherine Morayati

As you can see, neither font is correct and I can't change that.

I also tweaked other settings and they seemed to work, so the config file is being picked up.

mbikovitsky:
The "correct" way is to look up the full name of the font, not just the face name. This fixes #351.

### DIFF
--- a/garglk/fontwin.c
+++ b/garglk/fontwin.c
@@ -51,7 +51,7 @@ int CALLBACK monofont(
     if (!(strlen(gli_conf_monofont)))
         return 0;
 
-    if (!find_font_file(lpelfe->elfLogFont.lfFaceName, filepath, sizeof filepath))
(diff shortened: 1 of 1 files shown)


cspiegel:
Am I correct in assuming this bug affects all font selection on Windows? If so, would you agree that this warrants a new release for Windows users?

mbikovitsky:
Yes, this affects all font selection on Windows, so a new release is indeed warranted.

Issue #358, opened 2019-11-23
Freetype problem hinders compile


AndreasDavour:
Simply put, what the heck is freetype.pc and why is it not detected? I do have the freetype packes, including devel packages with headers, installed.

[ante@powerpad garglk]$ jam
OS is LINUX (gtk+) 
BUILD is RELEASE 
...patience...
...found 1816 target(s)...
...updating 505 target(s)...
Cc build/linux.release/garglk/main.o 
Package freetype was not found in the pkg-config search path.
Perhaps you should add the directory containing `freetype.pc'
to the PKG_CONFIG_PATH environment variable
Package 'freetype', required by 'virtual:world', not found
Archive build/linux.release/garglk/libgarglkmain.a 
ar: creating build/linux.release/garglk/libgarglkmain.a
Ranlib build/linux.release/garglk/libgarglkmain.a 
Cc build/linux.release/tads/argize.o 
Cc build/linux.release/tads/bif.o 
^C...interrupted
...updated 3 target(s)...
[ante@powerpad garglk]$ 
[ante@powerpad garglk]$ rpm -qa | grep freetype
freetype-2.9.1-7.fc30.x86_64
freetype-2.9.1-7.fc30.i686
freetype-devel-2.9.1-7.fc30.x86_64

cspiegel:
`.pc` files, along with the `pkg-config` program, are the more-or-less standardized way on Unix platforms to find headers and libraries.  If a `.pc` file isn't found, it typically means that the package that provides it isn't installed.

However, FreeType's `.pc` file is really called `freetype2.pc`, and as far as I can see all the calls to `pkg-config` in the Jam files are asking for `freetype2`, which is to say, I'm not at all sure where `freetype.pc` is coming from for you.

Can you run `jam -dx` and paste the output here, at least the output around the failing bit?  That will display which commands are run, and may help figure out which part of the build is asking for `freetype` vs `freetype2`.

AndreasDavour:
Thanks. I was not aware of the .pc files before. Good to know. 

That was when I had changes a bunch of things around due to other dependencies, but maybe it's better do run from the start. I did a clean clone and ran "jam -dx" and this is how it looks.

[garglk]$ jam -dx
OS is LINUX (gtk+) 
BUILD is RELEASE 

[...]

PR #360, opened 2019-12-16
Integrate rpm and deb build scripts.

interactivefiction:
Simple script to create an RPM build for gargoyle.  Tested on Fedora 29/30/31.


JamrulesRPM
```diff
@@ -0,0 +1,313 @@
+# You will probably need to edit this file to get everything to compile.
```
cspiegel: It looks like the only difference in this `Jamrules` file is that `USESDL` is disabled.  It'd be cleaner to use the original `Jamrules`, disabling `USESDL` on the command line, i.e.:

```
jam -sUSESDL=no
```
during the build step.  That way this file won't have to be kept in sync with the original.

gargoyle-buildrpm.sh
```diff
@@ -0,0 +1,150 @@
+#!/bin/sh
```
cspiegel: This script uses some non-POSIX stuff (e.g. `pushd`), so this should be `/bin/bash`.

gargoyle-buildrpm.sh
```diff
@@ -0,0 +1,150 @@
...
+
+if [ -d "./build/dist" ]
+then
+    rm -rfv ./build/dist
+fi
+if [ -d "./build/linux.release" ]
+then
+    rm -rfv ./build/linux.release
+fi
+#rm -rfv ./build/dist
```
cspiegel: These commented lines can be removed.

cspiegel:
I don't have any experience with RPM, but I built and installed RPM from source and gave this script a shot, and it wound up successfully building an RPM file for me.  I can't test it, but obviously you have.  There are a few comments I have, and I may wind up with a couple more things after looking this over again, but I'm happy merging it in when it's ready.

gargoyle-buildrpm.sh
interactivefiction: Will do.

JamrulesRPM
interactivefiction: Thanks for this, very helpful, will make the change!

interactivefiction: I attempted a build using the commandline approach you mention but the build system does not seem to honor the commandline flag.  Still getting SDL errors in the build.  Any thoughts?

gargoyle-buildrpm.sh
interactivefiction: Checked in changes.

gargoyle-buildrpm.sh
interactivefiction: Checked in changes.

JamrulesRPM
interactivefiction: Checked in changes.

interactivefiction:
Added a script to build DEB packages.  Script uses only dpkg-* tools, no higher level wrappers etc.  Tested resulting DEB packages on Ubuntu 16.04, 18.04, and 19.10.

cspiegel:
I'm attaching an RPM I created if you'd like to test; but it installs `/opt/rpm` (because my RPM is installed there), just as a note.

It looks like the script still disables SDL; in light of the conclusions of #358, should the scripts be updated to re-enable SDL?

interactivefiction:
I have a VM I can try and install that RPM to.  Out of curiosity, what environment are you building it in?  Yes, the script has SDL disabled only because I did the script work before I figured out the necessary RPM deps for SDL :->  I still need to build RPM packages with SDL enabled and put them through testing on fresh OS for F29,30,31 to make certain all of the new SDL runtime dependencies are properly accounted for during RPM install.  Currently, I do not know if the automatic dep generation properly accounts for them.  It should in theory, but testing it to be sure is probably best.  As the script is, I do know for sure that RPMs built without SDL will indeed install and auto-magically grab the required deps to run even if they are not currently installed on the system.

cspiegel:
> Out of curiosity, what environment are you building it in?

I'm on a custom Linux distribution, so there's no real useful way to ensure you have a compatible setup.

If you update the scripts to include SDL, I think there's no reason not to get them merged in.

interactivefiction:
Sounds good, when I get some spare time I will test with SDL on various systems and let you know.

cspiegel:
@interactivefiction A lot has changed in Gargoyle since you wrote these, the major change being the build system moving to CMake from Jam.

I rebased these changes onto the latest master and made some fixes to get them working with the changes.  I tested on Ubuntu 20.04 and Fedora 34.

Because I had to rebase onto a much later version, instead of trying to incorporate the fixes back into your branch, I created a new branch off of yours, rebased, and made a new pull request based on it: #540 

If you have any input regarding these new scripts, please let me know: I've never worked with either Debian or RPM build scripts before, so it's entirely possible I'm doing things in a less-than-ideal way.

cspiegel:
Closing after merging #540.  Thank you!

PR #322, opened 2018-05-18
fix: macOS Retina scaling issues (#320 and #321)

hkleinsorgen:
This fixes two scaling issues on macOS (#320 and #321). Resizing windows or moving them to displays with a different DPI setting now work as expected.

Issue #321, opened 2018-05-17
Text double size on standard DPI screen


angstsmurf:
With the new HiDPI changes, text will be will be double size on a standard DPI screen. When dragging from a standard DPI screen to a HiDPI or the other way, the text will change size. Only tested on macOS.

hkleinsorgen:
I have a patch for this issue and #320, which are closely related. Needs some cleanup, though.
On Windows I haven't observed such issues so far (except for the problem with the status bar).

angstsmurf:
That is great! Please consider making a pull request with the WIP code, so that it doesn't get lost.

angstsmurf:
As per my comment on #320, this is fixed by  #322.

PR #364, opened 2020-01-03
(Do not merge) Hi-DPI, rebased

angstsmurf:
This is just a reminder that we still have a Hi-DPI implementation that is actually in pretty good shape.

Unfortunately it now looks as if I wrote parts of this code rather than @hkleinsorgen and @townba. Perhaps there is some Git magic that can rectify this? There is nothing new from me included here.

All the bugs also remain. They are most apparent in games which implement a split screen with a graphics window on top, such as [Andromeda 1983](https://ifdb.tads.org/viewgame?id=k7gryitw24knktcc). Something seems to go awry when calculating the horizontal split window height. EDIT: Fixed by b747731.

Instead of this:
Skärmavbild 2020-01-03 kl  19 56 00
we get this:
Skärmavbild 2020-01-03 kl  19 53 40
The weird grey artefacts appear when resizing the window by dragging the corner.

### DIFF
--- a/garglk/config.cpp
+++ b/garglk/config.cpp
@@ -38,7 +38,7 @@
 bool gli_utf8input = true;
 bool gli_utf8output = true;
 
-int gli_hires = 1;
+bool gli_hires = true;
 float gli_backingscalefactor = 1.0f;
 float gli_zoom = 1.0f;
 
@@ -494,6 +494,10 @@ static void readoneconfig(const std::string &fname, const std::string &argv0, co
             gli_conf_sound = std::stoi(arg);
         } else if (cmd == "fullscreen") {
             gli_conf_fullscreen = std::stoi(arg);
+        } else if (cmd == "hires") {
+            gli_hires = std::stoi(arg);
+        } else if (cmd == "zoom") {
+            gli_zoom = std::stoi(arg);
         } else if (cmd == "speak") {
             gli_conf_speak = std::stoi(arg);
         } else if (cmd == "speak_input") {
--- a/garglk/config.cpp
+++ b/garglk/config.cpp
@@ -38,6 +38,10 @@
 bool gli_utf8input = true;
 bool gli_utf8output = true;
 
+int gli_hires = 1;
+float gli_backingscalefactor = 1.0f;
+float gli_zoom = 1.0f;
+
 struct gli_font_files gli_conf_prop, gli_conf_mono, gli_conf_prop_override, gli_conf_mono_override;
 
 std::string gli_conf_monofont = "Gargoyle Mono";
@@ -120,6 +124,7 @@ int gli_more_font = PROPB;
 unsigned char gli_scroll_bg[3] = { 0xb0, 0xb0, 0xb0 };
 unsigned char gli_scroll_fg[3] = { 0x80, 0x80, 0x80 };
 int gli_scroll_width = 0;
+int gli_scroll_width_save = 8;
 
 int gli_caret_shape = 2;
 int gli_link_style = 1;
@@ -468,6 +473,7 @@ static void readoneconfig(const std::string &fname, const std::string &argv0, co
             gli_link_style = !!std::stoi(arg);
         } else if (cmd == "scrollwidth") {
             gli_scroll_width = std::stoi(arg);
+            gli_scroll_width_save = gli_scroll_width;
         } else if (cmd == "scrollbg") {
             parsecolor(arg, gli_scroll_bg);
         } else if (cmd == "scrollfg") {
@@ -603,6 +609,24 @@ void gli_startup(int argc, char *argv[])
 
     if (!gli_baseline)
         gli_baseline = gli_conf_propsize + 0.5;
+
+   if (gli_hires)
+        gli_zoom *= gli_backingscalefactor;
+    gli_baseline = gli_zoom_int(gli_baseline);
+    gli_conf_monosize = gli_conf_monosize * gli_zoom;
+    gli_conf_propsize = gli_conf_propsize * gli_zoom;
+    gli_leading = gli_zoom_int(gli_leading);
+    gli_scroll_width_save = gli_zoom_int(gli_scroll_width_save);
+    gli_tmarginx = gli_zoom_int(gli_tmarginx);
+    gli_tmarginy = gli_zoom_int(gli_tmarginy);
+    gli_wborderx = gli_zoom_int(gli_wborderx);
+    gli_wbordery = gli_zoom_int(gli_wbordery);
+    gli_wmarginx = gli_zoom_int(gli_wmarginx);
+    gli_wmarginx_save = gli_zoom_int(gli_wmarginx_save);
+    gli_wmarginy = gli_zoom_int(gli_wmarginy);
+    gli_wmarginy_save = gli_zoom_int(gli_wmarginy_save);
+    gli_wpaddingx = gli_zoom_int(gli_wpaddingx);
+    gli_wpaddingy = gli_zoom_int(gli_wpaddingy);
 
     gli_initialize_tts();
     if (gli_conf_speak)
--- a/garglk/config.cpp
+++ b/garglk/config.cpp
@@ -497,7 +497,7 @@ static void readoneconfig(const std::string &fname, const std::string &argv0, co
         } else if (cmd == "hires") {
             gli_hires = std::stoi(arg);
         } else if (cmd == "zoom") {
-            gli_zoom = std::stoi(arg);
+            gli_zoom = std::stof(arg);
         } else if (cmd == "speak") {
             gli_conf_speak = std::stoi(arg);
         } else if (cmd == "speak_input") {
--- a/garglk/garglk.h
+++ b/garglk/garglk.h
@@ -118,6 +118,8 @@ typedef struct window_graphics_s window_graphics_t;
 #define HISTORYLEN 100
 
 #define GLI_SUBPIX 8
+#define gli_zoom_int(x) ((x) * gli_zoom + 0.5)
+#define gli_unzoom_int(x) ((x) / gli_zoom + 0.5)
 
 extern char gli_program_name[256];
 extern char gli_program_info[256];
@@ -237,6 +239,10 @@ extern int gli_wpaddingy;
 extern int gli_tmarginx;
 extern int gli_tmarginy;
 
+extern int gli_hires;
+extern float gli_backingscalefactor;
+extern float gli_zoom;
+
 extern bool gli_conf_lcd;
 extern unsigned char gli_conf_lcd_weights[5];
 
@@ -270,6 +276,7 @@ extern bool gli_conf_lockrows;
 extern unsigned char gli_scroll_bg[3];
 extern unsigned char gli_scroll_fg[3];
 extern int gli_scroll_width;
+extern int gli_scroll_width_save;
 
 extern int gli_baseline;
 extern int gli_leading;
--- a/garglk/garglk.h
+++ b/garglk/garglk.h
@@ -243,7 +243,7 @@ extern int gli_wpaddingy;
 extern int gli_tmarginx;
 extern int gli_tmarginy;
 
-extern int gli_hires;
+extern bool gli_hires;
 extern float gli_backingscalefactor;
 extern float gli_zoom;
 
--- a/garglk/garglk.h
+++ b/garglk/garglk.h
@@ -117,6 +117,10 @@ typedef struct window_graphics_s window_graphics_t;
 #define SCROLLBACK 512
 #define HISTORYLEN 100
 
+/* limit number of text rows/columns */
+#define MAX_TEXT_COLUMNS 255
+#define MAX_TEXT_ROWS 255
+
 #define GLI_SUBPIX 8
 #define gli_zoom_int(x) ((x) * gli_zoom + 0.5)
 #define gli_unzoom_int(x) ((x) / gli_zoom + 0.5)
--- a/garglk/garglk.h
+++ b/garglk/garglk.h
@@ -117,10 +117,6 @@ typedef struct window_graphics_s window_graphics_t;
 #define SCROLLBACK 512
 #define HISTORYLEN 100
 
-/* limit number of text rows/columns */
-#define MAX_TEXT_COLUMNS 255
-#define MAX_TEXT_ROWS 255
-
 #define GLI_SUBPIX 8
 #define gli_zoom_int(x) ((x) * gli_zoom + 0.5)
 #define gli_unzoom_int(x) ((x) / gli_zoom + 0.5)
--- a/garglk/garglk.ini
+++ b/garglk/garglk.ini
@@ -44,10 +44,7 @@ sound         1               # enable sound
 
 fullscreen    0               # set to 1 for fullscreen
 
-hires         1               # set to 1 to enable Retina display (macOS) / HiDPI (Windows)
-zoom          1.0             # set display zoom
-
-hires         1               # set to 1 to enable Retina display (macOS) / HiDPI (Windows)
+hires         1               # set to 1 to enable Retina display on macOS
 zoom          1.0             # set display zoom
 
 
--- a/garglk/garglk.ini
+++ b/garglk/garglk.ini
@@ -42,7 +42,13 @@ caps          0               # Force uppercase input  -- 0=off 1=on
 graphics      1               # enable graphics
 sound         1               # enable sound
 
-fullscreen    0               # set to 1 for fullscreen (Windows/Unix only)
+fullscreen    0               # set to 1 for fullscreen
+
+hires         1               # set to 1 to enable Retina display (macOS) / HiDPI (Windows)
+zoom          1.0             # set display zoom
+
+hires         1               # set to 1 to enable Retina display (macOS) / HiDPI (Windows)
+zoom          1.0             # set display zoom
 
 
 #===============================================================================
--- a/garglk/sysmac.mm
+++ b/garglk/sysmac.mm
@@ -363,6 +363,13 @@ void wininit(int *argc, char **argv)
 {
     NSAutoreleasePool * pool = [[NSAutoreleasePool alloc] init];
 
+    /* get the backing scale factor */
+    NSRect rect = NSMakeRect(0, 0, 1000, 1);
+    NSView * tmpview = [[NSView alloc] initWithFrame: rect];
+    rect = [tmpview convertRectToBacking: rect];
+    [tmpview release];
+    gli_backingscalefactor = NSWidth(rect)/1000.;
+
     /* establish link to launcher */
     NSString * linkName = [NSString stringWithUTF8String: getenv("GargoyleApp")];
     NSConnection * link = [NSConnection connectionWithRegisteredName: linkName host: NULL];
@@ -402,7 +409,9 @@ void winopen(void)
 
     [gargoyle initWindow: processID
                    width: defw
-                  height: defh];
+                  height: defh
+                  retina: gli_hires
+              fullscreen: gli_conf_fullscreen];
 
     wintitle();
     winresize();
@@ -528,7 +537,7 @@ void winkey(NSEvent *evt)
 
 void winmouse(NSEvent *evt)
 {
-    NSPoint coords = [evt locationInWindow];
+    NSPoint coords = [gargoyle getWindowPoint: processID forEvent: evt];
 
     int x = coords.x;
     int y = gli_image_h - coords.y;
--- a/garglk/sysmac.mm
+++ b/garglk/sysmac.mm
@@ -304,12 +304,14 @@ void winresize(void)
 {
     NSRect viewRect = [gargoyle getWindowSize: processID];
 
-    if (gli_image_w == (unsigned int) viewRect.size.width
-            && gli_image_h == (unsigned int) viewRect.size.height)
+    unsigned int vw = (unsigned int) (NSWidth(viewRect) * gli_backingscalefactor);
+    unsigned int vh = (unsigned int) (NSHeight(viewRect) * gli_backingscalefactor);
+
+    if (gli_image_w == vw && gli_image_h == vh)
         return;
 
-    gli_image_w = (unsigned int) viewRect.size.width;
-    gli_image_h = (unsigned int) viewRect.size.height;
+    gli_image_w = vw;
+    gli_image_h = vh;
     gli_image_s = ((gli_image_w * 4 + 3) / 4) * 4;
 
     /* initialize offline bitmap store */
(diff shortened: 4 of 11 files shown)


angstsmurf:
Rebased and working again. I wasn't sure what to do with the Windows-specific stuff in the recently removed `syswin.c`, so that is all gone for now.

garglk/config.cpp
```diff
@@ -38,6 +38,10 @@
 bool gli_utf8input = true;
 bool gli_utf8output = true;
 
+int gli_hires = 1;
```
cspiegel, line 41: This should be `bool` and `true` now (when these patches were originally done, `bool` didn't exist in the code).

garglk/garglk.ini
```diff
@@ -42,7 +42,13 @@ caps          0               # Force uppercase input  -- 0=off 1=on
...
 sound         1               # enable sound
 
-fullscreen    0               # set to 1 for fullscreen (Windows/Unix only)
+fullscreen    0               # set to 1 for fullscreen
+
+hires         1               # set to 1 to enable Retina display (macOS) / HiDPI (Windows)
+zoom          1.0             # set display zoom
+
+hires         1               # set to 1 to enable Retina display (macOS) / HiDPI (Windows)
+zoom          1.0             # set display zoom
```
cspiegel, line 51: These config options are listed twice.

Since the Windows bit has been removed, should this now just refer to macOS?

Qt has support for hidpi (https://doc.qt.io/qt-5/highdpi.html) but I don't have any clue how it works with Gargoyle on Windows (or Unix, for that matter), since I don't have a hidpi display.  It'd be nice to get tested on Windows/Unix and update this, but I'm also OK with a Mac-only fix for the time being, given that this is a known issue on Mac, and I've not seen a single bug report regarding Windows and hidpi.

garglk/window.c
```diff
@@ -602,8 +630,8 @@ void glk_window_get_size(window_t *win, glui32 *width, glui32 *height)
             hgt = hgt / gli_cellh;
             break;
         case wintype_Graphics:
-            wid = win->bbox.x1 - win->bbox.x0;
-            hgt = win->bbox.y1 - win->bbox.y0;
+            wid = gli_unzoom_int((win->bbox.x1 - win->bbox.x0));
```
cspiegel, line 633: Extra set of parentheses.

cspiegel (changes requested):

cspiegel:
> Rebased and working again. I wasn't sure what to do with the Windows-specific stuff in the recently removed `syswin.c`, so that is all gone for now.

As alluded to in a comment above, we can probably just not worry about Windows for the time being.  As long as things run fine on other platforms, then there's no reason to not get this in.  Thanks for keeping at this for so long.

garglk/garglk.ini
```diff
@@ -42,7 +42,13 @@ caps          0               # Force uppercase input  -- 0=off 1=on
 graphics      1               # enable graphics
 sound         1               # enable sound
 
-fullscreen    0               # set to 1 for fullscreen (Windows/Unix only)
+fullscreen    0               # set to 1 for fullscreen
+
+hires         1               # set to 1 to enable Retina display (macOS) / HiDPI (Windows)
+zoom          1.0             # set display zoom
```
cspiegel: These don't look to actually be configurable, i.e. there's no code in `config.cpp` to set them from the config file.  The hard-coded values are always used.

garglk/config.cpp
angstsmurf: Note to self: Build before pushing.

garglk/garglk.ini
curiousdannii: In Windows 10 you can set any screen to have a higher pixel density, by changing the display scale. It gets reported through to web browsers, so I assume Qt will identify the pixel density too.

garglk/config.cpp
```diff
@@ -38,6 +38,10 @@
...
+float gli_backingscalefactor = 1.0f;
```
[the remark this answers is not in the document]
angstsmurf: I didn't write this code and I don't pretend to understand it, but it seems that the gli_backingscalefactor variable is calculated at lines 368-374 in sysmac.mm, in the wininit method. It is a way to convert size in points to actual pixels. I think it will always be 2 on a Retina screen and 1 on a low resolution screen.

garglk/sysmac.mm
```diff
@@ -363,6 +365,13 @@ void wininit(int *argc, char **argv)
 {
     NSAutoreleasePool * pool = [[NSAutoreleasePool alloc] init];
 
+    /* get the backing scale factor */
+    NSRect rect = NSMakeRect(0, 0, 1000, 1);
+    NSView * tmpview = [[NSView alloc] initWithFrame: rect];
+    rect = [tmpview convertRectToBacking: rect];
+    [tmpview release];
+    gli_backingscalefactor = NSWidth(rect)/1000.;
+
```
angstsmurf, line 374: gli_backingscalefactor is set here.

garglk/config.cpp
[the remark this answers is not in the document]
curiousdannii: Common pixel ratios include 1, 1.5, 2, 3, but can also be 1.33, 1.67, 2.46, 2.61, 2.75... though those probably only occur on phones. Windows will let you enter any number (as a percentage) between 100 and 500, so someone might be running Gargoyle on something totally weird.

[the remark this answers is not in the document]
curiousdannii: Aside: I raised the idea of Glk functions to expose the pixel ratio here: https://github.com/erkyrath/glk-dev/issues/3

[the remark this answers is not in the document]
cspiegel: @angstsmurf Somehow I missed that assignment in sysmac.mm; thanks.

@curiousdannii I know nothing about hi res displays, so I'll be asking questions from ignorance here; would the Windows pixel ratio be used to calculate the backing scale on Windows?  Do you have a hidpi display on Windows, and if so, how does Gargoyle look on it?

If it's easy to get Windows hidpi support then it's worth putting it in, but unfortunately, I don't think I'd have a way to properly test it.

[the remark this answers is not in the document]
curiousdannii: It just looks blurred, as it's just upscaling from 100%.

You should be able to test it by changing one of your monitor's scales in the display settings. You don't need an actual hidpi screen.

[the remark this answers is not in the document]
hkleinsorgen: The gli_backingscalefactor for Windows  was computed in syswin.c:

https://github.com/garglk/garglk/commit/0e0aeeef6c331f298bb44838e8ac972f9642db6a#diff-dcdea0104f827a9d6043e4e1066cde7a21d77ead3c2d972dec57dcea055d1120L282

HiDPI support makes a big difference visually. I'd like to invest a few hours to see if Qt+HiDPI is possible on Windows an Linux. But maybe it's just me who is picky about the blurry fonts ;)

@angstsmurf : do you have a list of games with Hi-DPI issues apart from Andromeda 1983, for testing purposes?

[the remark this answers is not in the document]
cspiegel: > You should be able to test it by changing one of your monitor's scales in the display settings. You don't need an actual hidpi screen.

I only have Windows in a VM so I'm not sure how well that'll work, but I can give it a go.

> HiDPI support makes a big difference visually. I'd like to invest a few hours to see if Qt+HiDPI is possible on Windows an Linux. But maybe it's just me who is picky about the blurry fonts ;)

Thanks for giving this a look!

cspiegel (approved):

cspiegel:
If it winds up that Qt HiDPI is a lot of work, I'm fine merging this then creating a new request for the Windows work.

@angstsmurf If @hkleinsorgen is able to get the Qt side working, are you OK with those changes being merged into your branch here, so we have a single pull request for this?

garglk/config.cpp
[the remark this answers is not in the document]
angstsmurf: @hkleinsorgen A good test is probably [graphwintest.gblorb](https://www.eblong.com/zarf/glulx/graphwintest.gblorb), which is specifically written to test graphics windows, which seems to be where the bugs are.

Andromeda 1983 uses the "Location Images" extension, and all games which uses this seem to have the same problem. Another example is [Wolfsmoon](https://ifdb.org/viewgame?id=npxjxi6n9oak7zo7).

angstsmurf:
Yes, of course.

townba:
Two things I found:

* Minor issue: `gli_zoom` is being set using `std::stoi` instead of `std::stof`. Really only relevant for fractional zoom which isn't that big of a deal, but it's a simple fix.
* Major issue: This doesn't work with a dual monitor setup on Mac when one monitor is a Retina display and the other is not (my WFH setup is a MacBook Pro with its Retina screen and an external 1920x1080 monitor). In fact, I can't come up with a `.garglkrc` configuration that gets it to work. Unplugging the external monitor fixes the problem as long as I restart Gargoyle.

townba:
Ugh. I should clarify what "doesn't work" means. *Regardless of the `hires` setting in `.garglkrc` and which monitor it starts on*, it's not really usable. The window appears to be what would normally be the correct size, but it's always displaying just the upper left quadrant of what it should, filling the window.

Disabling `hires` (the default) has the font the correct size; enabling it makes the text twice as large as it should be. In both cases, games think the window is twice as wide and tall as it really is. The text isn't hi-res regardless of which monitor I start Gargoyle on unless I disconnect the non-Retina monitor.

Oof. Does that make sense?

garglk/config.cpp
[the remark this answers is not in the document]
hkleinsorgen: I'm facing some difficulties building garglk for Windows, neither cross compiling nor MSYS2 worked for me so far. So I suggest that you go ahead with macOS only for the time being.

hkleinsorgen:
> 
> 
> Ugh. I should clarify what "doesn't work" means. _Regardless of the `hires` setting in `.garglkrc` and which monitor it starts on_, it's not really usable. The window appears to be what would normally be the correct size, but it's always displaying just the upper left quadrant of what it should, filling the window.
> 
> Disabling `hires` (the default) has the font the correct size; enabling it makes the text twice as large as it should be. In both cases, games think the window is twice as wide and tall as it really is. The text isn't hi-res regardless of which monitor I start Gargoyle on unless I disconnect the non-Retina monitor.

Unfortunately, that sounds similar to the issues https://github.com/garglk/garglk/issues/321 and https://github.com/garglk/garglk/pull/322 that are supposed to be fixed :/

angstsmurf:
Sounds like at the very least the scale factor should recalculated when the window moves to a new screen. We already listen for a NSWindowDidChangeScreenNotification which calls [performRefresh:](https://github.com/garglk/garglk/blob/master/garglk/launchmac.mm#L478-L489), but I don't understand what performRefresh even does. Apparently not checking the resolution of the current screen.

hkleinsorgen:
The idea back then was to adapt to the window DPI on-the-fly when drawing things:
https://github.com/angstsmurf/garglk/blob/5cc1f63c6831d19ec4782a2fc855f9ec84f3fea5/garglk/launchmac.mm#L126

Maybe it wasn't a good idea...

angstsmurf:
No, that sounds like the sensible way to do it.

I don't understand most of this code and I don't have a way to test the multi-monitor stuff, so I'm really just guessing.

angstsmurf:
> There are three different coordinate systems:

So what you are saying is that this should work in theory and that you see nothing obviously wrong with the current code?

I should probably just stay out of this at least until I get a working cable for my second screen, but I'll just mention that #321 is not the same as the problem @hkleinsorgen has. The issue #321 was the text being fine on the hires screen, but becoming double size when the window was moved to the lores screen. [The current problem](https://github.com/garglk/garglk/pull/364#issuecomment-967775405) sounds more like the scaling is wrong (double) no matter what screen the window is on.

Also, I wonder if any of these are relevant. From [Apple's documentation on backingScaleFactor](https://developer.apple.com/documentation/appkit/nswindow/1419459-backingscalefactor?language=objc):

>Note
>For almost all common cases, developers should avoid using the value of backingScaleFactor as an input to layout or drawing calculations. Developers should instead use the backing coordinate space conversion methods, because the resulting code will more likely work consistently and correctly under both standard and high-resolution operation.

From a[ Stackoverflow answer](https://stackoverflow.com/a/54487166/10334672):

>The right thing for an OpenGL or Vulkan app to do is to render to the backing store resolution, not the physical pixel resolution.

angstsmurf:
I found my DisplayPort cable and tried this out with my exernal lo-res monitor. Strangely, my problem is the opposite of @hkleinsorgen's: With `hires 1` in the garglk.ini settings, everything works as it should: hi-res and sharp on my retina screen, lo-res and blurry (really blurry, with weird color bleed artifacts) on the external screen. But with `hires 0`, the text becomes half-sized and unreadably small on both screens.

angstsmurf:
… and the same is true with the external monitor disconnected. I apparently never tested this PR with the `hires` setting off before. When it is off, rather than making the text low resolution, it just shrinks it to half size. If `hires 0` is combined with `zoom 2`, the output is identical to `hires 1` and `zoom 1`.

garglk/config.cpp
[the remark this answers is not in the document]
cspiegel: @hkleinsorgen The Windows build is, to put it mildly, idiosyncratic vis-à-vis normal Windows development.  It expects a Unix-like environment with system packages that can be found via CMake, and probably pkg-config as well.

When I have some time I can see about making Windows support easier to handle, but I'm not sure I want to go back to bundling libraries as Gargoyle used to do.  Using existing packages is much cleaner, but if there are ways to be more inclusive (e.g. working with MSYS2, something I'm unfamiliar with) I'm all for it.

If you have an interest in a working Windows build, and don't mind a bit of setup work, you can use Gargoyle's MinGW Github Action as a guide.  It makes use of a prebuilt MinGW environment.  Setup is involved, but once it's done, it should be smooth.  Rough steps:

1. Install an Ubuntu 20.04 VM (this matches what GH Actions provide)
2. Install the following packages: build-essential cmake g++-mingw-w64 gcc-mingw-w64 mingw-w64-tools wget
3. Run `update-alternatives --set i686-w64-mingw32-gcc /usr/bin/i686-w64-mingw32-gcc-posix` and `update-alternatives --set i686-w64-mingw32-g++ /usr/bin/i686-w64-mingw32-g++-posix`
4. `wget https://github.com/garglk/assets/raw/master/mingw/mingw-i686.tar.xz && tar xfv mingw-i686.tar.xz -C /`
5. Run `NO_INSTALLER=1 ./windows.sh` from the top level of the `garglk` repository

At this point, if all goes well, you have `build-mingw` as the CMake build directory, and `build/dist` as the install target.  The script _should_ copy all necessary DLLs over, so from now on, all you have to do is run `make install` in the `build-mingw` directory to be able to quickly test the new changes, directly in `build/dist`.

I realize this is completely annoying, but once steps 1-4 are done, Windows development should go pretty smoothly.

hkleinsorgen:
Is switching to Qt on macOS an option?

cspiegel:
> Is switching to Qt on macOS an option?

Ultimately that'd be my preference.  I'd much prefer having only one interface to maintain.  However, the macOS interface operates differently than the other interfaces: the launcher remains open and in communication with the interpreter, and provides the ability to open new games from the File menu (so you can have multiple interpreters running simultaneously in the same session).

That's something that probably could be done with the Qt interface, but would require some significant changes in how it operates.  Qt already works on macOS, but it operates more like Unix or Windows: you have a single interpreter window, and when you close that, the whole program shuts down.  This is in contrast to what Mac users expect, which is that the program doesn't shut down till you actively quit via the file menu or a shortcut.  In all honesty Gargoyle isn't something that _needs_ the traditional Mac behavior, but previous Gargoyle releases worked that way, so it'd be a regression to go back.

cspiegel:
Speaking of the Qt build, I'm unable to create a bundle using it, because not all Qt libraries are being pulled in (it seems like QtDBus isn't being found, but that's just a preliminary look).  If a Qt bundle were buildable, then we could at least get decent comparisons, and I could include a Qt-based DMG during the CI process.  But I don't know how to deal with shared libraries on macOS like that, so if anybody wants to give it a try, it's just a matter of modifying the Mac build script to build against Qt6, and setting the minimum macOS version to 10.15 (required by Qt).

angstsmurf:
Simply removing the lines in window.c that zoomed graphic window seems to have fixed the problems in *Andromeda 1983* and *Wolfsmoon*, unless I'm missing something. There must have been another problem that this was meant to fix, but removing it seems to have no bad side effects.

There is still the garbage drawn "behind" the image when the window is manually resized, in the parts of the graphics windows that the image does not cover.

angstsmurf:
For me, the only remaining issue (except for the graphics window glitch mentioned above) is that text will be tiny with `hires 0`. I would suggest simply removing the `hires` option, which is how most applications work: always hires on a hires screen, always standard res on a standard dpi screen.

But @townba's report [above](https://github.com/garglk/garglk/pull/364#issuecomment-967773943) seems to suggest things are still broken on his multiple monitor setup even at `hires 1` and `zoom 1.0`. Is that correct?

townba:
I believe that was correct. I don't know when I'll be able to check it
again, though. Maybe a day or two.

garglk/garglk.h
```diff
@@ -117,7 +117,13 @@ typedef struct window_graphics_s window_graphics_t;
 #define SCROLLBACK 512
 #define HISTORYLEN 100
 
+/* limit number of text rows/columns */
+#define MAX_TEXT_COLUMNS 255
+#define MAX_TEXT_ROWS 255
```
cspiegel, line 122: These are only used in `window.c` so they ought to appear there so as not to pollute the global namespace.

What buffers are overflowed by having > 255 columns/rows?  Is there any reason not to just bump the size of those buffers (which would still require maximums of course), or switch them to dynamic buffers (if they're in C++ code, at least; dealing with dynamic buffers in C is annoying).

angstsmurf: This change is originally from #323. I don't know if problem it is meant to fix still exists.

cspiegel: OK, it looks like the problem exists (or at least _a_ problem with buffer overflows in that area exists).  If the viewport is larger than the scrollback buffer size then there are problems.  Switching to a dynamic scrollback buffer will ultimately be the right way to go but that's a lot of work.  Hard limits like this are fine for now, so just moving these constants into `window.c` works for the time being.

angstsmurf: Constants moved.

angstsmurf:
I found a way to trigger double size on my multi-screen setup with `hidpi 1`, but the latest change seems to have fixed it. I uploaded a [test binary](https://github.com/angstsmurf/garglk/releases/download/petter-mac-20211212/Gargoyle.zip) if anyone wants to try it out. `hidpi 0` is still broken.

erkyrath:
The test binary works out of the box on my Mac. (With no garglk.ini, that is.)

angstsmurf:
Thanks for testing!

cspiegel:
@angstsmurf I'm fine dropping the "hires" option entirely if the default works everywhere.

But on that note, I just tried the provided binary in a VM.  I'm running in low resolution mode (1920x1080) and I'm seeing the old "quarter screen" issue:
gargoyle-fs8
I guess this is because the plist file removed the "no hires" option, for obvious reasons.  My recollection is that this can be fixed on low-res screens, manually, by choosing to run it in low-res mode.  So the current tradeoff is: proper sizing for everybody, but blurry text on hires, or crisp hires text but quarter-size for low-res (but fixable).

Does that about sum it up?  If so I'm OK with prioritizing hires, given that it apparently looks _awful_ in low-res mode, and I figure a good chunk of Mac users are using hires monitors.  Unless I'm totally wrong here, I think we're about OK to merge.

Do you want me to do a PR for removing the hires option, or do you want to?  I don't mind doing it.

angstsmurf:
That's disappointing. I thought the VM I tested it in was standard resolution,, but apparently not. At least it sits in a different corner now, and the red background is gone.

I don't want this to be a release blocker, so I'm voting for just leaving things the way they were. Possibly make a separate beta / experimental release for Mac hires users only.

hkleinsorgen:
I encountered the same issue and created a fork with some additional fixes / refactorings:
https://github.com/hkleinsorgen/garglk/tree/hidpi-fixes

Compiled app for testing: https://drive.google.com/file/d/1aNzobFjcL8CKl07ySCo4XCISjYkADfpw/view?usp=sharing

There is some issue with fonts in the compiled app which I can't explain. The bold font in the graphics test is displayed as a regular font. The Andromeda adventures does not start due to a missing font. Theses issues were already present when compiling the original app, so I hope it's just an issue with my setup.

Changes:
- Does not use gli_hires, always use best resolution
- Removed some dead code
- Backing scale factor is recomputed whenever the window size changes.

The graphics / window test now looks ok when starting in lo-res, hi-res and scaled hi-res

Remaining issues
- The garbled graphics when resizing the window (probably just a bug that can be fixed without major changes)
- When changing the resolution, the zoom factor won't be updated
- Multiple host windows on screens with different resolution won't work

Zoom and backing scale are unfortunately globals. Moving them to the Gargoyle window structure would be help fixing the remaining issues. Another non-trivial issue is that fonts are initialized once at startup. Fonts would need to be reinitialized whenever the zoom / backing scale factor for a host window changes.

angstsmurf:
Thanks a lot for working on this!

I'm not entirely sure if this is what you mean by "multiple host windows on screens with different resolution won't work", but with your test app, the Gargoyle window is alway double size on my standard res secondary monitor. Moving the window to my MacBook retina screen makes it normal size again.

It does not seem to matter which monitor I start Gargoyle on or what settings I change. The exception is setting the secondary monitor to mirror the laptop screen (rather than expanding it), and then to "More space", which I guess means native 1:1 resolution: Gargoyle's resolution will then change as intended.

hkleinsorgen:
Another test app:  https://drive.google.com/file/d/1V5rGf7YKWGvsWRIaEx2KlOxgphuYsr9k/view?usp=sharing
Source: https://github.com/hkleinsorgen/garglk/tree/hidpi-static-backing

This uses a static backing factor of 2.0. This solves the issues when changing the resolution, and it should work with multi-monitor setups with different resolutions (I can't test that, though).

The downsides:
- unnecessary resource usage on low res displays
- might be more blurry on low res displays due to the additional scaling

angstsmurf:
This one works!

erkyrath:
Looks good to me. (Note that I no longer have a low-res Mac to test on.)

hkleinsorgen:
Low-res modes can be set with a hidden feature of the monitor settings:
https://www.howtogeek.com/415159/how-to-select-an-exact-display-resolution-on-your-mac/

erkyrath:
I don't think those prefs let you change the display from double-resolution to single-resolution. They change the logical dot pitch. (This may indirectly affect the 2x behavior but it's not obvious how.)

cspiegel:
That one's working for me at 1920x1080 (again, in a VM, but I have to assume that's a reasonable facsimile of a real 1920x1080 monitor).  The text doesn't look particularly blurry, either.

hkleinsorgen:
> I don't think those prefs let you change the display from double-resolution to single-resolution. They change the logical dot pitch. (This may indirectly affect the 2x behavior but it's not obvious how.)

You need to check the "[ ] Show low resolution modes" checkbox and then select a resolution marked as "low resolution" from the list. Choosing other resolutions has the effect you described.

hkleinsorgen:
@cspiegel @angstsmurf 
How to proceed? Merging my fork of angstsmurf's fork of an older commit might be difficult.
Should I create a new merge request based on the current main branch?
Or should angssmurf merge my changes?

angstsmurf:
I'll just close this and let you make a new one. Excellent work!

PR #367, opened 2020-02-04
Autocomplete with Tab

ricpelo:
This patch brings word autocompleting typing the first characters of a word and pressing Tab.

I'm not the original author. Sorry but I don't know who is. I found it on a forum several years ago. After some tweaking here and there, I think it's working very good.

### DIFF
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -191,7 +191,7 @@ static void loadglyph(font_t *f, glui32 cid)
     int adv;
 
     gid = FT_Get_Char_Index(f->face, cid);
-    if (gid == 0)
+    if (gid <= 0)
         gid = FT_Get_Char_Index(f->face, '?');
 
     for (x = 0; x < GLI_SUBPIX; x++)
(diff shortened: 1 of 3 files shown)


cspiegel:
I know you didn't write the initial code so you may not have some of the answers, but I'm going to make some comments about this patch that we'll need to get resolved one way or another.

I'm willing to get a tab completion patch in, but it'll very likely have to be optional (though it could probably default to being on).

ricpelo:
> I know you didn't write the initial code so you may not have some of the answers, but I'm going to make some comments about this patch that we'll need to get resolved one way or another.
> 
> I'm willing to get a tab completion patch in, but it'll very likely have to be optional (though it could probably default to being on).

Feel free to do anything you need, of course. Hope this PR helps!

cspiegel:
Apologies, but apparently a review I started when this was first posted has been pending this whole time.  I want to blame the GitHub UI, but it's mainly my fault for not paying close attention, I suppose.

cspiegel:
I'm closing this for now just because the approach needs work, and needs somebody dedicated to taking it on.  If non-intrusive, full tab completion support is made available, I have no problem incorporating it.

I'll probably eventually take a look myself if nobody else does, but it'll be after the next release.

PR #369, opened 2020-02-17
Remove Frotz and Nitfol.

cspiegel:
Although Frotz is maintained, the maintained version does not include
Glk support. The Frotz that has been included in Gargoyle until now was
patched specifically to add Glk support for Gargoyle. Keeping it in sync
with upstream would require a lot of effort.

Nitfol has been unmaintained for years.


curiousdannii:
Will this alter the level of version 6 support, or did Glk-Frotz have the same sort of limitations as Bocfel?

cspiegel:
> Will this alter the level of version 6 support, or did Glk-Frotz have the same sort of limitations as Bocfel?

Bocfel has been the default version 6 interpreter since the 2011 release, so there's no alteration there.  But none of the interpreters has good V6 support anyway: Glk is incompatible with how V6 does windows so it's never going to be a particularly great way to play these games.  I do have a local branch of Bocfel which "supports" pictures (by just dumping them at the cursor position), which I'll probably wind up merging in at some point, but something like Zoom will always be better.

DavidGriffith:
I've been looking at Glk as an official option (https://gitlab.com/DavidGriffith/frotz/issues/109).  Somehow someone got Glk and Frotz to work together with @erkyrath's https://github.com/erkyrath/remglk.  The ScummVM project (https://github.com/scummvm/scummvm/) also has Frotz working with Glk for IO and it does a decent job with V6.  Maybe that IO interface layer will work?

cspiegel:
The ScummVM interpreters have been hacked to the point that the [ScummVM Glk Wiki](https://wiki.scummvm.org/index.php?title=Glk) says, “don't anticipate that it'll be easy to take any interpreters converted specifically for ScummVM Glk and simply drop them into any other Glk implementations.”

For all upstream interpreters, it's ideal for them to conform to the Glk spec, possibly with Gargoyle-specific extensions.  The best is just being able to plug in an interpreter without touching the code at all.  If Frotz gets a Glk IO interface, it'd be a perfect candidate, although I can understand if that's not a priority.

DavidGriffith:
I'll get something going to add Glk support to Frotz.

PR #378, opened 2020-05-14
Allow passing a game to gargoyle's .desktop entry

Beuc:
This allows e.g. dragging a game to the toolbar icon, or associating a file extension with gargoyle in Gnome.

See:
https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=882768


cspiegel:
Well that fixes a very overlooked but very annoying bug.  Thanks!

PR #382, opened 2020-09-08
Improved Bocfel sound support

angstsmurf:
Adds support for callbacks to properly fade and loop sound effects in Sherlock and The Lurking Horror. Basically I've copied and pasted code from Frotz, so it might not be super clean and consistent, but it seems to work.

Main caveats: only tested on macOS. Seems to play samples faster (at double speed?) than the almost identical code in Spatterlight and Frotz. Is this due to a difference between SDL 1 and SDL 2?

The latest Bocfel seems to require an extra ZTERP_GLK_BLORB flag to build with Blorb support.

EDIT: A simple way to test this is to type >BLOW WHISTLE, WAIT at the beginning of Sherlock: The Riddle of the Crown Jewels. Requires the right version of the game and the .blb version of the sounds, of course.

### DIFF
--- a/terps/Jamfile
+++ b/terps/Jamfile
@@ -129,7 +129,7 @@ if $(MAKE_BOCFEL) = yes
     SubDir TOP terps bocfel ;
     ObjDir bocfel ;
 
-    SubDirCcFlags -std=c99 -DZTERP_GLK -DGARGLK ;
+    SubDirCcFlags -std=c99 -DZTERP_GLK -DGARGLK -DZTERP_GLK_BLORB ;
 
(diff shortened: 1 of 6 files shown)


cspiegel:
Inspired by this merge request I've added support for sound callbacks in Bocfel 1.3.  I did some fairly large changes around how interrupt routines are handled, so had to do this in a different manner, by and large.

The Lurking Horror hack still isn't done, and I'm not sure if I will add it or not.  Probably at some point, but I wanted to prioritize the non-hacky behavior first, so at least fading works now.

PR #388, opened 2020-10-07
Additional sound formats used by Hugo games

angstsmurf:
This adds support for the MOD formats used by the Hugo games Enceladus and Dragon Hunt, and the WAVE format used by Hugo Tetris. It also tries to add support for General MIDI, but because the only game that uses it, Crimson Spring, won't start in Gargoyle (and never has, as far as I can tell) it is as yet untested. 

Use of the apparently unsafe tempnam() function is removed here as well.

EDIT: I have only tested this on macOS, and it would be good if someone tried it on Windows.

Enceladus can be downloaded here: https://ifdb.tads.org/viewgame?id=lwqcpp5od2vmslg
And Dragon Hunt here: https://ifdb.tads.org/viewgame?id=htl5pcglo0yjs0cy

### DIFF
--- a/garglk/sndsdl.c
+++ b/garglk/sndsdl.c
@@ -662,11 +662,10 @@ static glui32 play_compressed(schanid_t chan, char *ext)
 }
 
 /** Start a mod music channel */
-static glui32 play_mod(schanid_t chan, long len, char *ext)
+static glui32 play_mod(schanid_t chan, long len)
 {
     FILE *file;
-    char *tn;
-    char *tempdir;
+    char *tn, *tempdir;
     int music_busy, loop;
 
     if (chan == NULL)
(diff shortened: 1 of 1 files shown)


garglk/sndsdl.c
```diff
@@ -672,9 +688,16 @@ static glui32 play_mod(schanid_t chan, long len)
 
     chan->status = CHANNEL_MUSIC;
     /* The fscking mikmod lib want to read the mod only from disk! */
-    tempdir = getenv("TEMP");
-    if (tempdir == NULL) tempdir = ".";
-    tn = tempnam(tempdir, "gargtmp");
+    tempdir = getenv("TMPDIR");
```
cspiegel, line 691: `getenv` can return NULL so this needs to take that into account.

garglk/sndsdl.c
```diff
@@ -672,9 +688,16 @@ static glui32 play_mod(schanid_t chan, long len)
...
+    /* malloc size of string tempdir + "XXXXXX.' + 3 letter extension + terminator */
+    tn = malloc(strlen(tempdir) + strlen("XXXXXX.mod") + 1);
+    sprintf(tn, "%sXXXXXX", tempdir);
+#if defined(WIN32) || defined(_WIN32) || defined(WINDOWS)
+    _mktemp_s(tn, strlen(tn));
+#else
+    mkstemp(tn);
```
cspiegel, line 698: The reason `tempnam` is insecure is due to a race condition: it creates a filename but can't guarantee an attacker won't create that file before the program does.  `mkstemp` is secure because it creates the file with a secure file mode and opens it simultaneously, returning the file descriptor, removing the chance for a race.

By taking the filename and tacking on an extension, the race comes back here, in that attackers could again create the file.  That's not really much of a concern, I'd think, with Gargoyle, but it does mean that the linker warning about `tempnam` being insecure is just hidden, not fixed.

Moreover, `mkstemp` returns a file descriptor, which is leaked here.

Is the extension actually necessary for SDL to determine the file format?  The existing code looks like it just passes a filename without extension, meaning SDL is capable of determining file format via file contents, not name.  If so, then there's no need for the extension here, and this can be made cleaner with something like:

```c
int fd;
fd = mkstemp(tn);
if (fd == -1) /* handle error */
file = fdopen(fd, "wb");
```
When `file` is closed, the underlying descriptor will be closed as well.

This also doesn't require separate code for Windows: MinGW is used to build the Windows binaries, so it's safe to assume a MinGW environment, which includes `mkstemp` and `fdopen`.

If SDL does require an extension, then it might be OK to use the non-standard `mkstemps` function, which allows a suffix to be added to the temporary filename.  This exists on at least Linux (glibc and musl), FreeBSD, NetBSD, and OpenBSD, although not MinGW, so that would need to use the racy version, but that's at least no worse than it used to be.

angstsmurf: Thanks! It should be better now.

PR #389, opened 2020-10-09
Add a case-insensitive string unput function.

cspiegel:


angstsmurf:
Is there a downside to making the unput functions always case-insensitive by default, rather than making it an optional flag? It would be a little slower, I suppose, but do you know of anything that would break?

cspiegel:
I considered doing that.  My concern would be for anybody using garglk outside of this repository, as this is a breaking change.  I would say that I don't expect many people actually are using it, and fewer still (probably 0) using the unput functions, so it may be a valid change.

There's no promise of stability in Gargoyle's Glk extensions, but there is some tracking of them done outside of Gargoyle (@curiousdannii & @cibersheep for example), so I'm a bit on the fence about this one, given that it's a low-impact change.

I can let this pull request sit for a while to see if there are, in fact, any interested parties who take issue with changing the behavior of the unput functions.  It definitely would make the change cleaner if I didn't have to pass around enums all over the place.

cibersheep:
I don't know what this would imply but if you are worried about my little project, I'm not using Garglk directly but remGLK... so we would be ok, I think :)

cspiegel:
I've decided to just make the unput functions case insensitive as suggested.

PR #391, opened 2020-10-25
Expose the garglk_fileref_get_name function with GLK_MODULE_FILEREF_GET_NAME

curiousdannii:
Expose the garglk_fileref_get_name function through GLK_MODULE_FILEREF_GET_NAME so other Glk implementations can add it.

I realised this was necessary while investigating why TADS couldn't save/restore in emglken/Lectrote.

https://intfiction.org/t/lectrote-1-3-8/47931/16?u=dannii

I also patched TADS's `os_askfile` function so that if GLK_MODULE_FILEREF_GET_NAME isn't defined it will just error, rather than saying it succeeded only for another function down the line to error. (This is instead of guarding the function call directly with `#ifdef GARGLK` as you did in cspiegel/terps.)

### DIFF
--- a/garglk/glk.h
+++ b/garglk/glk.h
@@ -57,6 +57,7 @@ typedef int32_t glsi32;
 #define GLK_MODULE_HYPERLINKS
 #define GLK_MODULE_DATETIME
 #define GLK_MODULE_GARGLKTEXT
+#define GLK_MODULE_FILEREF_GET_NAME
 
 /* These types are opaque object identifiers. They're pointers to opaque
     C structures, which are defined differently by each library. */
--- a/tads/osglk.c
+++ b/tads/osglk.c
@@ -520,6 +520,10 @@ int os_askfile(const char *prompt, char *fname_buf, int fname_buf_len,
     frefid_t fileref;
     glui32 gprompt, gusage;
 
+#ifndef GLK_MODULE_FILEREF_GET_NAME
+    return OS_AFE_FAILURE;
+#endif
+
     if (prompt_type == OS_AFP_OPEN)
         gprompt = filemode_Read;
     else
--- a/terps/agility/os_glk.c
+++ b/terps/agility/os_glk.c
@@ -5922,7 +5922,7 @@ gagt_get_user_file (glui32 usage, glui32 fmode, const char *fdtype)
    * underlying file descriptor or FILE* from a Glk stream either. :-(
    */
 
-#ifdef GARGLK
+#ifdef GLK_MODULE_FILEREF_GET_NAME
   retfile = fopen(garglk_fileref_get_name(fileref), fdtype);
 #else
 
--- a/terps/alan2/exe.c
+++ b/terps/alan2/exe.c
@@ -1526,7 +1526,16 @@ void save()
   char str[256];
   AtrElem *atr;
 
-#ifndef GARGLK
+#ifdef GLK_MODULE_FILEREF_GET_NAME
+
+  frefid_t fref;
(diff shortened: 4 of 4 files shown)


cspiegel:
On a side note, I saw your related post on intfiction.org and created an exploratory issue in the terps repository at cspiegel/terps#5, if you've got any feedback there.  Since that repository is actually being used, it ought to be a bit better maintained.

Thanks for the patch!  I'll update Bocfel to support the new macro as well.

curiousdannii:
I'm planning to switch to using this repo as it's kept up to date more. There's no real downside (for me) to using a submodule of this repo as I don't use the CMake files the other repo provides.

curiousdannii:
https://github.com/erkyrath/remglk/pull/36

Zarf thinks the better option is to add a function to glkstart.h called `glkunix_fileref_get_filename`. What do you think we should do? Rewrite all uses of `garglk_fileref_get_name` in this repo? Use `#define` to map between the two names? I'd appreciate your thoughts especially in regards to using this repo (or terps) with other Glk implementations.

cspiegel:
I'm not sure I follow the rationale for using a `glkunix_` prefix vs a `garglk_` prefix for non-standard functionality/extensions to Glk. It seems that `glkunix_` ought to be for Unix-specific stuff, although the current functions aren't Unix-specific anyhow (e.g. `glkunix_stream_open_pathname` is at least something that could exist on Win32).  I suppose an argument can be made that Glk implementations could exist places where the concept of a filename is meaningless, whereas something like `garglk_set_zcolors` is at least a plausible function across all Glk implementations, thus putting it in a place of "universal" extensions, so to speak.  I think I may have convinced myself while writing this.

But I don't have strong feelings one way or the other, so if it makes sense to put this in `glkstart.h` as `glkunix_fileref_get_name`, I'll merge a pull request that does that, barring any strong objections.  If that's the case, I think all uses should just be changed, given that this is not a public API and as such is only "officially" used by Gargoyle.  We can excise `garglk_fileref_get_name` entirely.

@angstsmurf : I know you're tracking some Gargoyle-specific functionality so I'm pinging you here for your input as well as to give you a heads up for potential changes you might want to make to Spatterlight.

curiousdannii:
Btw, I think I can just add a `#define` to map the function names in emglken, so you don't have to make any changes here. It's more about what we think is best for the whole Glk ecosystem. Which I really don't know.

When GarGlk compiles Glulxe on Windows, does it use unixstrt.c? If so that does kinda prove that it's not really UNIX specific... 

And on the other hand, winstart.c calls `winglk_fileref_create_by_name` and `glk_stream_open_file`, which is mostly just a more awkward way of calling `glkunix_stream_open_pathname`.

cspiegel:
> When GarGlk compiles Glulxe on Windows, does it use unixstrt.c? If so that does kinda prove that it's not really UNIX specific...

It's treated as a Unix implementation regardless of platform, so it does expose the glkunix functions no matter what.

Zarf would be the authority but it looks like the separate Unix code is more to support Unix conventions (command-line arg parsing) and then non-windowed file manipulation (`glkunix_stream_open_pathname` as opposed to an "Open file" dialog), not because they're *only* useable on Unix.  But the concept of needing a filename does seem to be more universal, given that WinGlk has `winglk_get_initial_filename`.

However, I will agree with Zarf that `garglk_fileref_get_name` is a hack.  It seems to be used in Gargoyle by interpreters that were ported to Glk and still use standard I/O routines, using Glk in this instance only to present an "Open file" dialog, discarding the Glk handle afterward.  From what I can tell, if Gargoyle provided an equivalent to `winglk_get_initial_filename`, then `garglk_fileref_get_name` wouldn't even be necessary.

PR #393, opened 2020-11-09
Be a lot more specific when searching for fonts

### LINKED ISSUE #392: Wrong fonts chosen [fontconfig]
When using fontconfig as the font backend (e.g. on Unix), Gargoyle can choose the wrong font files.

As a concrete example, take the font [Source Code Pro](https://github.com/adobe-fonts/source-code-pro).  Set this as `monofont` in the config file (`monofont Source Code Pro`).

Grab [TerpETude](https://ifarchive.org/if-archive/infocom/interpreters/tools/etude.tar.Z).

Run TerpETude through Gargoyle, and select option 4 (Styled text).  Observe the following:

fixed-fs8

“Italic” looks the same weight as “Bold”, for example.  This is not determinstic, so it may work on other systems, or it may choose different but incorrect fonts.  But the main problem is that when Gargoyle asks for a particular style, fontconfig returns all matching fonts in some random-appearing order, from which Gargoyle picks one.

Fonts which only have 4 styles (Regular, Italic, Bold, Bold Italic) probably won't exhibit this, but fonts with many different styles can.

cspiegel:
Previous to this, Gargoyle was not as strict in what it asked for
regarding fonts. As an example, when asking for the font:

Source Code Pro:style=Regular

On my system, the following fonts were returned:

/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-Regular.otf
/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-Semibold.otf
/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-Black.otf
/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-ExtraLight.otf
/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-Light.otf
/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-Medium.otf

Gargoyle would just choose the first in the list. As it happens the
first here was wanted. However, for the font:

Source Code Pro:style=Italic

The following comes back:

/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-BlackIt.otf
/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-LightIt.otf
/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-MediumIt.otf
/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-It.otf
/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-ExtraLightIt.otf
/usr/share/fonts/truetype/sourcecodepro/SourceCodePro-SemiboldIt.otf

The first entry here is Black Italic, which is not what is desired, but
which does match the requested style: it is Italic, after all.

Now the code is a lot more specific in what it asks for. For example,
instead of asking just for Italic, it asks for Regular Italic (along
with related variations such as Book Italic), which avoids the wrong
weights.

This fixes #392.

### DIFF
--- a/garglk/fontfc.c
+++ b/garglk/fontfc.c
@@ -21,7 +21,6 @@
  *****************************************************************************/
 
 #include <fontconfig/fontconfig.h>
-#include <stdarg.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -30,6 +29,8 @@
 #include "glk.h"
 #include "garglk.h"
 
+static bool initialized = false;
+
 static bool findfont(const char *fontname, char *fontpath, size_t n)
 {
     FcPattern *p = NULL;
@@ -37,9 +38,6 @@ static bool findfont(const char *fontname, char *fontpath, size_t n)
     FcObjectSet *attr = NULL;
     FcFontSet *fs = NULL;
     bool success = false;
-
-    if (!FcInit())
-        return false;
 
     p = FcNameParse((FcChar8*)fontname);
     if (p == NULL)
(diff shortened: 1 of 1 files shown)


cspiegel:
I don't know if this is at all the correct approach, but I can't find any useful documentation on fontconfig at all, so this involved trial and error as well as digging through the fontconfig source itself.

PR #397, opened 2020-12-21
Bocfel: Fix display of negative numbers in status bar

angstsmurf:
In at least the z3 versions of Zork I, a score of -10 is displayed as 65256 in Bocfel (and Yazmin and Lectrote, but not in Frotz.)

In release 75 and later, you can get a score of -10 by typing
```
> open mailbox, take leaflet, throw leaflet at me
```
as you first move.

Reported by user Morningstar in [this thread](https://intfiction.org/t/mac-voiceover-users-wanted-for-testing-again/48653/4).

I don't know if this is the best fix or what regressions it may cause.

### DIFF
--- a/terps/bocfel/screen.c
+++ b/terps/bocfel/screen.c
@@ -1953,7 +1953,7 @@ void zshow_status(void)
 #ifdef ZTERP_GLK
   glui32 width, height;
   char rhs[64];
-  int first = variable(0x11), second = variable(0x12);
+  int16_t first = variable(0x11), second = variable(0x12);
 
   if(statuswin.id == NULL) return;
 


cspiegel:
To be perfectly pedantic converting from `uint16_t` to `int16_t` should use the `as_signed()` function, since it's implementation-defined behavior to convert from an out-of-range value to a signed integer.  Of course it'd be a perverse platform to not perform the conversion as we'd expect...

As with the Cutthroats time issue, this should also be done in `/status`, but I'll make that change locally here and bring it in before the next release.  I think there'll be a new Bocfel release soonish, as I'm finalizing a pretty large internal change here, and plan to do a release after I've merge it in.  That may well be enough reason to push out another Gargoyle release as there have been several Bocfel bugs/issues fixed, thanks to you.

I'd say this is an obviously correct change, since the standard says “The score may be assumed to be in the range -99 to 999 inclusive”, meaning it must be a signed value.

cspiegel:
I'm attaching a patch representing the changes I've made locally for this.

The significant introduction I've made is using `long` instead of `int` to hold the values.  In Bocfel I do my best to adhere strictly to C99, and `int` won't necessarily hold all values `int16_t` can (it's valid to have an `int` which holds only values from -32767 to 32767, whereas `int16_t` is guaranteed to be 2's complement, so -32768 to 32767).  So `first` and `second` should be at least `int16_t`. But since they're used in `printf()`, promotion rules apply, meaning `%d` might not be proper, and the macros for the fixed-width integers are awful, so `long` it is.

In all reality any system which provides `int16_t` will also have an `int` which is at least as large.  If it were a real pain to conform (unnecessarily, admittedly) to the standard I wouldn't, but this is pretty painless.
[status.txt](https://github.com/garglk/garglk/files/5738179/status.txt)

angstsmurf:
Thanks! I've applied this to the Spatterlight source now.

angstsmurf:
I hope you won't mind if I close these two.

PR #398, opened 2020-12-22
Bocfel: Fix time display quirk in Cutthroats

angstsmurf:
If you drop your watch (can be done at the first move) the status bar should display Time: 99:99 pm. The original Amiga interpreter doesn't, but that is probably a bug.

For explanation and discussion, see:
https://intfiction.org/t/invalid-values-for-time-in-v3-games/43928

### DIFF
--- a/terps/bocfel/screen.c
+++ b/terps/bocfel/screen.c
@@ -1977,8 +1977,18 @@ void zshow_status(void)
 
   if(status_is_time())
   {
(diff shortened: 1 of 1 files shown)


cspiegel:
I've got a local branch of Bocfel that does this that I'll probably merge in before the next release.  There's the slight wrinkle that this ought to also be implemented for `/status` as well.  It's not hard to do at all, but a bit annoying since the two are in separate files.

Issue #380, opened 2020-08-05
depends on deprecated GTK 2


Beuc:
Hi,

I received this information at the Debian bug tracker, I thought I'd relay the info:

This package has Build-Depends on GTK 2 (libgtk2.0-dev), or produces
binary packages with a Depends on GTK 2.

GTK 2 was superseded by GTK 3 in 2011 (see
<https://bugs.debian.org/947713>). It no longer receives any significant
upstream maintenance, and in particular does not get feature development
for new features like UI scaling on high-pixel-density displays (HiDPI)
and native Wayland support. GTK 3 is in maintenance mode and GTK 4 is
approaching release, so it seems like a good time to be thinking about
minimizing the amount of GTK 2 in the archive.

GTK 2 is used by some important productivity applications like GIMP, and
has also historically been a popular UI toolkit for proprietary software
that we can't change, so perhaps removing GTK 2 from Debian will never be
feasible. However, it has reached the point where a dependency on it is
a bug - not a release-critical bug, and not a bug that can necessarily
be fixed quickly, but a piece of technical debt that maintainers should
be aware of.

A porting guide is provided in the GTK 3 documentation:
https://developer.gnome.org/gtk3/stable/migrating.html

[...]
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=967369

cspiegel:
This is definitely an issue that needs to be addressed.  I've been toying around for a while with a Qt version which I have working in a broken fashion.  If necessary I'll eventually try a port to Gtk3, but if I can get Qt working instead that'd be ideal.

In short, this does need to happen, but I don't have a timeframe for it.

cibersheep:
> [...] a Qt version which I have working in a broken fashion.

Something we can see ^__^?
I'm developing a similar project for Ubuntu Touch (which uses Qt) using remGLK. I don't know if any of the dirty code could help you but: https://gitlab.com/cibersheep/gelek-vanilla

The interesting bit is I managed to build the terps (old versions maybe) for arm64, armhf and amd64 architectures using [clickable](https://clickable-ut.dev/)

wyatt8740:
GTK2 still works, for now at least. I don't really think that's a problem.

Supporting additional newer toolkits could still be a great benefit, however. I personally haven't been able to grok Qt yet (C++ tends to hurt my small brain), but from the 'user perspective' it does have a nicer file picker GUI and seems less prone to drastically changing every few months.

Just my two cents.

[...]

PR #400, opened 2021-01-09
Port to gtk3

mstoeckl:
This patch ports the GTK2 frontend to GTK3, closing #380.

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -32,7 +32,7 @@ jobs:
         fuse
         git
         grep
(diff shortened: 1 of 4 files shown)


cspiegel:
I've verified that this works for Unix and Windows, but I can't test OS X.  Has this been tested there to ensure it still works (not just compiles, but runs properly)?  If not, I'd appreciate anybody with the ability to build on OS X to give it a try, then this can be merged in.

angstsmurf:
Seems to be working here on macOS 10.15.7.

PR #402, opened 2021-01-30
Port to SDL2

### LINKED ISSUE #376: Port to SDL2
Gargoyle currently uses SDL1.  It should be updated to use SDL2 instead.

mstoeckl:
This closes #376.

Testing that I've been able to do:
* Building on Linux; using Soundtest.gblorb to verify that mod, ogg, wav formats still play as before. I could not find a file with an embedded mp3 and thus was unable to check mp3 support directly.
* Cross compiling for Windows with `jam CROSS=1 MINGWARCH=i686-w64-mingw32` worked, and I was able to run gargoyle under Wine. Was not able to test sound due to the way I was using Wine. (Note: the build with x86_64-w64-mingw32 was failing even before this patch.)


angstsmurf:
I can confirm that this works on macOS if the garglk/Jamfile is edited (removing `|| $(OS) = MACOSX` from https://github.com/garglk/garglk/blob/master/garglk/Jamfile#L104 is enough, but I guess the rest of the statement should go as well if it is going to build on Solaris.)

mstoeckl:
Updated to remove that statement with SDL_sound.

cspiegel:
Just one note here: the include directives for SDL in `garglk/sndsdl.c` shouldn't include the `SDL2/` prefix: the pkg-config files that come with SDL already include it (e.g. `pkg-config sdl2 --cflags` → `-I/usr/include/SDL2`).  It works if SDL is installed in a standard search path, e.g. `/usr/include`, but not if it's somewhere non-standard, so:

```diff
diff --git a/garglk/sndsdl.c b/garglk/sndsdl.c
index a648a22..fca45e1 100644
--- a/garglk/sndsdl.c
+++ b/garglk/sndsdl.c
@@ -29,8 +29,8 @@
 #include <assert.h>
 #include <math.h>
 
-#include <SDL2/SDL.h>
-#include <SDL2/SDL_mixer.h>
+#include <SDL.h>
+#include <SDL_mixer.h>
 
 #include "glk.h"
 #include "garglk.h"
```
Past that, it looks good to me on both Linux and Windows (sound is working just fine on Windows when tested via Wine).

mstoeckl:
Updated again, to fix those two #includes.

mstoeckl:
Would you mind making a new release to include this sometime in the next 6 months or so?

cspiegel:
I think a release within 6 months is very doable.  I've got a couple largeish changes to Bocfel that I'll want to get in first but that should probably happen within a couple weeks.

Issue #399, opened 2021-01-03
Bocfel: Initial slash in input breaks Beyond Zork definitions menu


angstsmurf:
Here is another way to break the Beyond Zork definitions menu: move to the beginning of a line, by deleting the preloaded input text or moving past it with the left arrow key, type a slash and enter. Bocfel will print a long help message about meta commands in the buffer window, and the input field will move to the left edge of the line below its proper position.

cspiegel:
This kind of thing also happens in the game _My Angel_ when in novel mode.  I fixed this locally a couple months ago after testing with _My Angel_, but it turns out to only have been a partial fix in light of this bug report: BZ works _better_ in my local branch, but due to the fact that it's using preloaded input, it's not perfect.  The preloaded text makes it a bit tricky, but I'm working on a fix.  Thanks for the report!

cspiegel:
@angstsmurf OK, a new Bocfel release has been merged in which fixes this issue.  And should also fix #384, #385, and #395.

If you'd like to double-check that all work for you, please do so, then the issues can be closed.  Otherwise I'll just close them, as my testing indicates they're now working.

angstsmurf:
Yay, it works. Excellent job!

angstsmurf:
Right, there is still a very small glitch: if you move the cursor all the way to the left, type a slash and hit enter, the command will be echoed with an extra vertical bar at the end, which is a bit fiddly to get rid of. But that is a nitpick.

cspiegel:
Interesting; it looks to be because there's nothing erasing the text that was previously there.  It can be more "obviously" recreated by typing in a very long, nonsense meta-command, like "/aaaaaaaaaaaaaaaaaaaaaa".

The fix is to overwrite the user's input with spaces.  It can't be completely guaranteed to be right in all cases, but it's at least better than leaving the text there.  And it seems to be fine for Beyond Zork.

Here's a patch that seems to work: [bz.txt](https://github.com/garglk/garglk/files/5820953/bz.txt)

angstsmurf:
That works great! Thanks for all your hard work.

cspiegel:
I appreciate the beating you're putting Bocfel through. All these weird edge cases would have gone undiscovered if not for your poking and prodding.

[...]

Issue #383, opened 2020-09-13
Bocfel: glk_window_move_cursor call sets wrong cursor position


angstsmurf:
The call to glk_window_move_cursor() in Bocfel at[ line 326 in screen.c](https://github.com/garglk/garglk/blob/master/terps/bocfel/screen.c#L326) seems to move the cursor, i.e. the current printing position in the status window, an extra `input->len` number of characters to the right. 
This breaks the Function Key Definitions functionality in Beyond Zork.

To reproduce, type DEFINE as a command in Beyond Zork. Then press any function key. Duplicate text is printed (and line input started) at the wrong position.

Removing the line seems to fix the problem without any obvious adverse effects.

EDIT: Actually, removing [line 318](https://github.com/garglk/garglk/blob/master/terps/bocfel/screen.c#L318) instead seems to be a better idea. After implementing mouse support, the other way caused some issues.

cspiegel:
Yeesh, this is an ugly one.  Simply removing the line doesn't work because the original code is correct in the case of no preloaded input _but_ input was terminated by a terminating key.  The problem is that preloaded text is printed, resulting in `input->x` being advanced already, thus the doubling.  It looks like the following change to line 318 works, if I'm understanding all the interactions properly:

```c
if(input->term != ZSCII_NEWLINE) upperwin->x += input->len - input->preloaded;
```

But you've also uncovered another problem, as the preloaded text in the DEFINE menu is broken (giving both a lower- and uppercase version of the text).  This is ridiculous because it used to work, but I fixed a similar problem for preloaded text at the command prompt: in previous versions, hitting a function key would result in something like:

`EXAMINEexamine`

Now in the DEFINE menu, you get:

`examineEXAMINE`

The cases are swapped, and I used the knowledge that BZ was preloading input with lowercase to hack around the problem.  But for the DEFINE menu it's preloading in uppercase, but printing lower!  The exact opposite problem, meaning the hack doesn't work there.

I think I'll just add a case-insensitive "unput" function to Gargoyle instead of trying to work around this inside of Bocfel.

angstsmurf:
Yes, that line fixes it. Thanks!

angstsmurf:
Closed by #401

PR #404, opened 2021-02-20
Update the unput functions to return the number of unput characters

cspiegel:
@angstsmurf This represents a change in the garglk API, so I'd like your input, since Spatterlight implements garglk extensions.  This is needed for a more complete fix for a Beyond Zork bug.  I can't find the actual issue (I thought it was #399 but maybe not); 

Anyway, you might recall that there was one fix that required the subtraction of the preloaded text length in order to get proper cursor placement.  I was going on the assumption that preloaded text would always match printed text (as is the case in BZ), but of course the Z-machine allows preloaded text to be different from anything on screen.  Instead of the length of preloaded text, the length of unput text needs to be counted, so I modified the unput functions to return a count of how many characters got removed.

If there's a better fix that doesn't involve changing the API I'm all for it, but so far this is what I've come up with.

### DIFF
--- a/garglk/cgstream.c
+++ b/garglk/cgstream.c
@@ -923,13 +923,13 @@ static void gli_put_buffer_uni(stream_t *str, glui32 *buf, glui32 len)
     }
 }
 
-static void gli_unput_buffer(stream_t *str, char *buf, glui32 len)
+static glui32 gli_unput_buffer(stream_t *str, const char *buf, glui32 len)
 {
     glui32 lx;
-    unsigned char *cx;
+    const unsigned char *cx;
 
     if (!str || !str->writable)
-        return;
+        return 0;
 
     if (str->type == strtype_Window)
     {
@@ -943,7 +943,7 @@ static void gli_unput_buffer(stream_t *str, char *buf, glui32 len)
             else
             {
                 gli_strict_warning("unput_buffer: window has pending line request");
-                return;
+                return 0;
             }
         }
         for (lx=0, cx=buf+len-1; lx<len; lx++, cx--)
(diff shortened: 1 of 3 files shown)


curiousdannii:
The safe way to handle the API change would be to introduce new functions, for example `gli_unput_buffer_returning_length`. The existing functions could then be wrappers around the new functions.

angstsmurf:
Is it #383 you're looking for?

How widely adopted is the old API? Do we know how many interpreters currently implement it?

Unless there are more than a couple, I think it is fine to just make this change. Every function we add will probably make adoption by other interpreter writers less likely.

cspiegel:
Yes, #383 is the one.

I'm on the fence about a new vs modified API here.  No interpreters in the Gargoyle tree, apart from Bocfel, even use the unput functions.  It's possible that somebody out there has written an interpreter that uses them, of course.

But there are at least two other Glk implementations that implement Gargoyle extensions (Spatterlight and Fabularium), so it's probably for the best to have new API functions instead of modifying these.

angstsmurf:
I'm happy either way.

cspiegel:
I'm in agreement that bloating the amount of Gargoyle extensions to Glk should be avoided, and in an ideal world I'd prefer to be able to modify it whenever I want, keeping the in-tree interpreters up to date.  But I'd be hypocritical if I did that, given that I developed Bocfel out-of-tree, and against Gargoyle.  It was nice for me to have stable extension APIs, and any silent developers out there should be given the same affordance.

But I'll always want to balance that against other Glk implementors who use extensions; it's best for us to be compatible as much as possible.

I've updated the request to add new functions.  I took the opportunity to make the parameters `const` as well.

PR #407, opened 2021-02-20
SNDSDL: Fix #396 MOD loop bug

### LINKED ISSUE #396: MOD music loop bug
I'll just report it here as I can't seem to fix it myself: When playing MOD music, the current SDL sound code in Gargoyle plays music once when it is told to play it twice, i.e. repetitions is set to 2. It works correctly for 1 and 3 repetitions, but fails for 2. (This may be confusing: remember that 1 repetition means play the sound once, 0 repetitions means play nothing at all etc.)

It works as it should when playing OGG or AIFF sounds. It also works correctly in Spatterlight with pretty much identical code, which makes me suspect that this is a bug in SDL 1, or rather in the Mikmod library that SDL 1 uses for MODs. (Spatterlight uses SDL 2.)

This can be tested with the [Sound Test](https://github.com/angstsmurf/soundtest/releases) program. Just >SET LOOPS TO 2 and >PLAY MOD. Compare with >LOOP 1. (I know that the default songs are on the long side, which quickly becomes annoying. I took the shortest loopable non-copyrighted sounds I could find.)

angstsmurf:
I haven't tested this extensively, and I was apparently confused about this before, so it would be great if somebody else could confirm that this works.

### DIFF
--- a/garglk/sndsdl.c
+++ b/garglk/sndsdl.c
@@ -635,7 +635,7 @@ static glui32 play_mod(schanid_t chan, long len)
         SDL_UnlockAudio();
         Mix_VolumeMusic(chan->volume);
         Mix_HookMusicFinished(&music_completion_callback);
-        loop = chan->loop - 1;
+        loop = chan->loop;
         if (loop < -1)
             loop = -1;
         if (Mix_PlayMusic(chan->music, loop) >= 0)


cspiegel:
This works for me now, and seems like the obviously-correct approach.

The [SDL documentation](https://www.libsdl.org/projects/SDL_mixer/docs/SDL_mixer.html#SEC57) gets this wrong.  It says, for the loops parameter: “0 plays the music zero times...”

But the SDL code is:

```c
if (loops == 0) {
    /* Loop is the number of times to play the audio */
    loops = 1;
}
```

So this bug was hidden as long as it was only tested with a loop value of 1.

PR #414, opened 2021-04-11
Upgrade Bocfel to version 1.3

cspiegel:


angstsmurf:
As you mentioned, the sound code isn't complete, so perhaps this is already known. 

The Lurking Horror will halt at the "Place" scene with `Fatal error: too many (253) locals at 0x1a (pc = 0x5d16)`, after playing the first sample and trying to loop.

To test: `turn on pc. login 872325412. password uhlersoth. click box. click paper. read paper. z.z.z.z`.

cspiegel:
@angstsmurf As always, you're tracking down my bugs.  Thank you!

I've pushed fixes for those two issues.

angstsmurf:
> I've pushed fixes for those two issues.

That's better!

I also tested the new mouse support in Beyond Zork a bit, and it seems to work fine.

angstsmurf:
"The Spy Who Came In From The Garden" still crashes with `Fatal error: too many (38) locals at 0x4 (pc = 0x4917)` if you ring the doorbell at the starting location.

There are many broken variants of this file on the web, so I attached a working version.

[spy.zblorb.zip](https://github.com/garglk/garglk/files/6299885/spy.zblorb.zip)

cspiegel:
I'm not sure what to do about “The Spy Who Came In From The Garden”.  I've already got a workaround in place which claims to be for that game (having a repeat value of 0 actually mean 1), but since the standard doesn't actually mention that game as the reason (I may have read that on raif or something, years ago), I'm going to drop the reference to it from the comment.

Anyway, the point here is that the call to `@sound_effect` in the attached version of "Spy" above is incorrect.  It's doing this:

```
SOUND_EFFECT    L00,#02,#ff,4
```

The value of 4 for the routine is invalid.  I'm not sure what the intent was here, but I'm not sure it's worth a workaround.  I would much rather have a _proper_ version of the game file available, although a cursory search hasn't turned up any Inform source code from which to make one.

I've hacked the file you attached so that it no longer includes a broken call.  In the absence of the original source being available, it'd be great if _this_ could be the canonical Blorb demo, since it at least doesn't crash.  It's available here: [new-spy.zip](https://github.com/garglk/garglk/files/6300463/new-spy.zip)

Workarounds for Infocom games are one thing, but for a perpetually-broken Blorb demo, I'm just not sure.  I suppose a patch could be added for it easily, so maybe I'll do that, instead of modifying the sound code.

I think I'm going to do a proper 1.3.1 release of Bocfel in a day or two with these fixes, plus proper repeats for The Lurking Horror: and let me tell you, figuring out the meaning/point of `lh_repeats` in the Frotz code was not particularly straightforward!  I'll keep this pull request open for now but I'll likely just close it out and replace it in the near future.  I might keep adding tentative 1.3.1 patches while it's open, though.

cspiegel:
On reflection, the [official version](https://inform-fiction.org/zmachine/standards/blorb/index.html#nine) of "Spy" is hopelessly broken, and, if I'm not mistaken, the version attached here is a one-off from [this intfiction.org post](https://intfiction.org/t/cblorb-versus-perlblorb/4341/11).  I think it's safe to just relegate "Spy" to the dustbin of history.  It's not worth adding a patch for.

angstsmurf:
> I think it's safe to just relegate "Spy" to the dustbin of history. It's not worth adding a patch for.

Fair enough. I wouldn't worry about it.

cspiegel:
Closing in favor of #415.

PR #415, opened 2021-04-15
Update Bocfel to 1.3.1.

cspiegel:


angstsmurf:
This looks great. I haven't really looked at the new autosave code, but I think it might be useful to improve the Spatterlight autosaving as well.

I dug deeper into the "queued" sound effects in The Lurking Horror, and it seems they are currently not working right in Frotz either. I made a [pull request](https://github.com/garglk/garglk/pull/416) (based on this one) to support them, but you can probably work out a better way to do it.

cspiegel:
Oh, that save file's great.  I actually have a local branch which theoretically supports the queued sounds, but I haven't been able to test it against LH yet, so I didn't include it in 1.3.1.  I'll look at it with the save file to see if it's done properly.

cspiegel:
And things turn out more complicated than they should be.  I don't think the hack about queueing sounds is quite enough, because LH produces something like this in very quick succession after “put line in socket”:

```
@sound_effect 11 2 8
@sound_effect 16 2 8
@sound_effect 16 3
```

So, start effect 11, then immediately start 16, then immediately stop 16.

If sounds are queued, it's probable that sound effect 11 will still be playing when the call to stop 16 comes in.  Then, later, 16 starts, with nothing to stop it.  16 is a looped effect (which, incidentally, isn't addressed by the standard), so after 11 finishes, 16 starts and doesn't stop.  You'd have to queue stop requests as well, which starts getting ugly.

So what I'm seeing in Frotz is the behavior described above, where sound 11 plays, then 16 plays perpetually, meaning the stop request is swallowed.

In Bocfel, a call to stop a sound effect will stop _any_ playing effect, not just the requested one, which is obviously not entirely correct, but unless Infocom meant to support mixing effects, a call to stop a sound should always match with the most recently-requested sound.

I see in your patch that you just ignore repeats on a queued sound, which is an elegant solution to the problem.  I've added that approach to my local branch, and I think things are working fine.

My tentative patch is here: [lh-hack.zip](https://github.com/garglk/garglk/files/6326776/lh-hack.zip)

angstsmurf:
The sounds seem to work great with the patch applied.

I tried the game with the original Amiga interpreter in an emulator, and it seems to cut the three longest of the four final samples a little short. Only the cracking shell is played to completion.

I don't know if the original Amiga disk drive this was written for was so slow that the sounds were always played to completion before it finished reading the next one from disk, or if this was simply the best they could do with the technical limitations.

Anyway, I'd say the way we do it in Bocfel now is sufficiently close to the original.

cspiegel:
Alright, last(?) one: #417 is an update to Bocfel 1.3.2, which includes the Lurking Horror queued sound hack.

With that, sound support is almost complete.  All that's missing is beeps, and I plan on adding a Gargoyle extension for that just as soon as I find appropriate (free) sound effects.

PR #417, opened 2021-04-17
Update Bocfel to 1.3.2

cspiegel:


angstsmurf:
For reference, I recorded a video with audio of The Lurking Horror in the emulator here:
https://youtu.be/C0TRiLe26Tw

cspiegel:
That's extremely interesting, thank you for doing it.  I got UAE working with The Lurking Horror a few days ago so I could reproduce earlier issues (right at the start of the game), but I didn't know how to get to the portion of the game requiring queued sound effects.  I thought it'd be a bit over the top to ask you if you'd be willing to play through it on a (slow!) emulator, so this was a pleasant surprise.

The time between turns really is excruciatingly slow, so it's understandable why the sound effects were done "wong" in the code, given that the sound-enabled game was on the Amiga only.

angstsmurf:
> The time between turns really is excruciatingly slow

Well, it is quite snappy on the Amiga compared to the Commodore 64 version.

Issue #418, opened 2021-04-21
Gamma correction is performed incorrectly when rendering text (especially light text on a dark background)


Tynach:
At first I thought the gamma correction was just being done incorrectly when the text was lighter than the background, but I've since been comparing the output of programs like `ftview` and `ftstring` to even dark-on-light text, and I'm seeing excessively blurry results in this program. I've tried to find if you're setting a custom filter weight to do this, and haven't found evidence that you are.

Instead I found a whole lot of very strange code in `draw.c` that looks like manual LCD filtering, instead of trusting FreeType to do this correctly.

I have no idea why you are doing it this way. Maybe I'm missing something (I do have a headache and haven't gone through all the code to figure out how it's all pieced together), or I might be misinterpreting what functions like `gammacopy_lcd()` are doing.

At the very least, gamma correction is *not* working for light-on-dark text. Instead it seems to be applying the inverse of the gamma (the light text gets thinner on dark backgrounds). In reality, gamma correction *should* have whichever one is lighter seem to have a more intense effect against whichever is darker, so dark text on a light background makes the text seem thinner, while light text on a dark background makes the text seem thicker.

cspiegel:
> I have no idea why you are doing it this way.

That makes two of us.  I'm not at all familiar with gamma correction and I've never looked at the relevant code here.  Your comment of “trusting FreeType” implies that it's able to do what Gargoyle is doing manually; is that correct?

I wouldn't have any problems ditching Gargoyle code in favor of FreeType, assuming the output is good/better, which it sounds liek it ought to be.  But again, this isn't my wheelhouse.  If you are able to provide some sample code, that would be very helpful.

Tynach:
> > I have no idea why you are doing it this way.
> 
> That makes two of us. I'm not at all familiar with gamma correction and I've never looked at the relevant code here. Your comment of “trusting FreeType” implies that it's able to do what Gargoyle is doing manually; is that correct?
> 
> I wouldn't have any problems ditching Gargoyle code in favor of FreeType, assuming the output is good/better, which it sounds liek it ought to be. But again, this isn't my wheelhouse. If you are able to provide some sample code, that would be very helpful.

You happened to narrow in on the one part of my post that I have determined to be a red herring. I was accidentally using a custom-built copy of the FreeType library that I happened to have in my library paths, and that was the source of the overly blurry text. Rebuilding Gargoyle to use the system copy of FreeType fixed that minor issue, but not the issue of incorrect gamma correction.

[...]

PR #419, opened 2021-05-07
Significantly improve the gamma correction for more natural looking text rendering

Tynach:
It actually does mean a lot to me that Gargoyle actually cares about providing gamma-correct text rendering for those of us who want that. I'm not sure of any other open source programs that advertise that as a feature, let alone implement it.

Sadly, Gargoyle's gamma correction code was implemented incorrectly, and as a result there were various inconsistencies and problems with text rendering. Most noticeable was when you'd have light text on a dark background, as that would make the text very thin and hard to make out, even leading to a comment in the config files suggesting that people with dark themes use a gamma value that's lower than 1.0.

This should no longer be necessary with this code, and I've indeed removed the relevant comments from those config files. More details about that, and many of the other changes, are in the commit messages for the changes I've made. Some information about how gamma correction in general works - along with some of my earlier musings before actually trying to implement the code myself - can be found in the comments to issue #418, which this set of commits completely fixes.

### DIFF
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -31,11 +31,12 @@
 #include <ft2build.h>
 #include FT_FREETYPE_H
 #include FT_OUTLINE_H
+#include FT_LCD_FILTER_H
 
 #include <math.h> /* for pow() */
 #include "uthash.h" /* for kerning cache */
 
-#define mul255(a,b) (((a) * ((b) + 1)) >> 8)
+#define mul255(a,b) (((a) * (b)) >> 8)
 #define grayscale(r,g,b) ((30 * (r) + 59 * (g) + 11 * (b)) / 100)
 
 #ifdef _WIN32
@@ -86,6 +87,8 @@ struct font_s
  */
 
 static unsigned char gammamap[256];
+static unsigned char gammainv[256];
+const unsigned char filterweights[5] = {28, 56, 85, 56, 28};
 
 static font_t gfont_table[8];
 
@@ -128,19 +131,18 @@ static int touni(int enc)
     return enc;
 }
 
-static void gammacopy(unsigned char *dst, unsigned char *src, int n)
+static void copy(unsigned char *dst, unsigned char *src, int n)
 {
     while (n--)
-        *dst++ = gammamap[*src++];
+        *dst++ = *src++;
 }
 
 #define m28(x) ((x * 28) / 255)
 #define m56(x) ((x * 56) / 255)
 #define m85(x) ((x * 85) / 255)
 
-static void gammacopy_lcd(unsigned char *dst, unsigned char *src, int w, int h, int pitch)
+static void copy_lcd(unsigned char *dst, unsigned char *src, int w, int h, int pitch)
 {
-    const unsigned char zero[3] = { 0, 0, 0 };
     unsigned char *dp, *sp;
     int x, y;
 
@@ -150,15 +152,9 @@ static void gammacopy_lcd(unsigned char *dst, unsigned char *src, int w, int h, 
         dp = &dst[y * pitch];
         for (x = 0; x < w; x += 3)
         {
-            const unsigned char *lf = x > 0 ? sp - 3 : zero;
-            const unsigned char *rt = x < w - 3 ? sp + 3 : zero;
-            unsigned char ct[3];
-            ct[0] = gammamap[sp[0]];
-            ct[1] = gammamap[sp[1]];
-            ct[2] = gammamap[sp[2]];
-            dp[0] = m28(lf[1]) + m56(lf[2]) + m85(ct[0]) + m56(ct[1]) + m28(ct[2]);
-            dp[1] = m28(lf[2]) + m56(ct[0]) + m85(ct[1]) + m56(ct[2]) + m28(rt[0]);
-            dp[2] = m28(ct[0]) + m56(ct[1]) + m85(ct[2]) + m56(rt[0]) + m28(rt[1]);
+            dp[0] = sp[0];
+            dp[1] = sp[1];
+            dp[2] = sp[2];
             sp += 3;
             dp += 3;
         }
@@ -212,10 +208,18 @@ static void loadglyph(font_t *f, glui32 cid)
         if (f->make_oblique)
             FT_Outline_Transform(&f->face->glyph->outline, &ftmat);
 
-        if (gli_conf_lcd)
+        if (gli_conf_lcd) {
+            if (gli_conf_gamma == 1.0) {
+                FT_Library_SetLcdFilterWeights(ftlib, filterweights);
+            } else {
+                FT_Library_SetLcdFilter(ftlib, FT_LCD_FILTER_LIGHT);
+            }
+
             err = FT_Render_Glyph(f->face->glyph, FT_RENDER_MODE_LCD);
-        else
+        } else {
             err = FT_Render_Glyph(f->face->glyph, FT_RENDER_MODE_LIGHT);
+        }
+
         if (err)
             winabort("FT_Render_Glyph");
 
@@ -229,13 +233,13 @@ static void loadglyph(font_t *f, glui32 cid)
         glyphs[x].data =
             malloc(glyphs[x].pitch * glyphs[x].h);
                 if (gli_conf_lcd)
-                    gammacopy_lcd(glyphs[x].data,
-                            f->face->glyph->bitmap.buffer,
-                            glyphs[x].w, glyphs[x].h, glyphs[x].pitch);
+                    copy_lcd(glyphs[x].data,
+                       f->face->glyph->bitmap.buffer,
+                       glyphs[x].w, glyphs[x].h, glyphs[x].pitch);
                 else
-                    gammacopy(glyphs[x].data,
-                            f->face->glyph->bitmap.buffer,
-                            glyphs[x].pitch * glyphs[x].h);
+                    copy(glyphs[x].data,
+                       f->face->glyph->bitmap.buffer,
+                       glyphs[x].pitch * glyphs[x].h);
     }
 
     if (cid < 256)
@@ -384,7 +388,10 @@ void gli_initialize_fonts(void)
     int i;
 
     for (i = 0; i < 256; i++)
-        gammamap[i] = pow(i / 255.0, gli_conf_gamma) * 255.0;
+        gammamap[i] = (unsigned char)(pow(i / 255.0, gli_conf_gamma) * 255.0 + 0.5);
+
+    for (i = 0; i < 256; i++)
+        gammainv[i] = (unsigned char)(pow(i / 255.0, 1.0 / gli_conf_gamma) * 255.0 + 0.5);
 
     err = FT_Init_FreeType(&ftlib);
     if (err)
@@ -426,18 +433,29 @@ void gli_draw_pixel(int x, int y, unsigned char alpha, unsigned char *rgb)
 {
     unsigned char *p = gli_image_rgb + y * gli_image_s + x * gli_bpp;
     unsigned char invalf = 255 - alpha;
+    unsigned char bg[3] = {
+        gammamap[p[0]],
+        gammamap[p[1]],
+        gammamap[p[2]]
+    };
+    unsigned char fg[3] = {
+        gammamap[rgb[0]],
+        gammamap[rgb[1]],
+        gammamap[rgb[2]]
+    };
+
     if (x < 0 || x >= gli_image_w)
         return;
     if (y < 0 || y >= gli_image_h)
         return;
 
 #ifdef EFL_1BPP
-    int gray = grayscale( rgb[0], rgb[1], rgb[2] );
-    p[0] = gray + mul255((short)p[0] - gray, invalf);
+    int gray = grayscale( fg[0], fg[1], fg[2] );
+    p[0] = gammainv[gray + mul255((short)bg[0] - gray, invalf)];
 #else
-    p[0] = rgb[2] + mul255((short)p[0] - rgb[2], invalf);
-    p[1] = rgb[1] + mul255((short)p[1] - rgb[1], invalf);
-    p[2] = rgb[0] + mul255((short)p[2] - rgb[0], invalf);
+    p[0] = gammainv[fg[2] + mul255((short)bg[0] - fg[2], invalf)];
+    p[1] = gammainv[fg[1] + mul255((short)bg[1] - fg[1], invalf)];
+    p[2] = gammainv[fg[0] + mul255((short)bg[2] - fg[0], invalf)];
 #ifndef WIN32
     p[3] = 0xFF;
 #endif
@@ -451,19 +469,30 @@ void gli_draw_pixel_lcd(int x, int y, unsigned char *alpha, unsigned char *rgb)
         invalf[0] = 255 - alpha[0];
         invalf[1] = 255 - alpha[1];
         invalf[2] = 255 - alpha[2];
+    unsigned char bg[3] = {
+        gammamap[p[0]],
+        gammamap[p[1]],
+        gammamap[p[2]]
+    };
+    unsigned char fg[3] = {
+        gammamap[rgb[0]],
+        gammamap[rgb[1]],
+        gammamap[rgb[2]]
+    };
+
     if (x < 0 || x >= gli_image_w)
         return;
     if (y < 0 || y >= gli_image_h)
         return;
 
 #ifdef EFL_1BPP
-    int gray = grayscale( rgb[0], rgb[1], rgb[2] );
+    int gray = grayscale( fg[0], fg[1], fg[2] );
     int invalfgray = grayscale( invalf[0], invalf[1], invalf[2] );
-    p[0] = gray + mul255((short)p[0] - gray, invalfgray);
+    p[0] = gammainv[gray + mul255((short)bg[0] - gray, invalfgray)];
 #else
-    p[0] = rgb[2] + mul255((short)p[0] - rgb[2], invalf[2]);
-    p[1] = rgb[1] + mul255((short)p[1] - rgb[1], invalf[1]);
-    p[2] = rgb[0] + mul255((short)p[2] - rgb[0], invalf[0]);
+    p[0] = gammainv[fg[2] + mul255((short)bg[0] - fg[2], invalf[2])];
+    p[1] = gammainv[fg[1] + mul255((short)bg[1] - fg[1], invalf[1])];
+    p[2] = gammainv[fg[0] + mul255((short)bg[2] - fg[0], invalf[0])];
 #ifndef WIN32
     p[3] = 0xFF;
 #endif
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -97,10 +97,10 @@ int gli_image_w = 0;
 int gli_image_h = 0;
 unsigned char *gli_image_rgb = NULL;
 
-#if defined(WIN32)
-static const int gli_bpp = 3;
-#elif defined EFL_1BPP
+#ifdef EFL_1BPP
 static const int gli_bpp = 1;
+#elifdef WIN32
+static const int gli_bpp = 3;
 #else
 static const int gli_bpp = 4;
 #endif
@@ -430,18 +430,17 @@ void gli_draw_pixel(int x, int y, unsigned char alpha, unsigned char *rgb)
         return;
     if (y < 0 || y >= gli_image_h)
         return;
-#ifdef WIN32
-    p[0] = rgb[2] + mul255((short)p[0] - rgb[2], invalf);
-    p[1] = rgb[1] + mul255((short)p[1] - rgb[1], invalf);
-    p[2] = rgb[0] + mul255((short)p[2] - rgb[0], invalf);
-#elif defined EFL_1BPP
+
+#ifdef EFL_1BPP
     int gray = grayscale( rgb[0], rgb[1], rgb[2] );
     p[0] = gray + mul255((short)p[0] - gray, invalf);
 #else
     p[0] = rgb[2] + mul255((short)p[0] - rgb[2], invalf);
     p[1] = rgb[1] + mul255((short)p[1] - rgb[1], invalf);
     p[2] = rgb[0] + mul255((short)p[2] - rgb[0], invalf);
+#ifndef WIN32
     p[3] = 0xFF;
+#endif
 #endif
 }
 
@@ -456,11 +455,8 @@ void gli_draw_pixel_lcd(int x, int y, unsigned char *alpha, unsigned char *rgb)
         return;
     if (y < 0 || y >= gli_image_h)
         return;
-#ifdef WIN32
-    p[0] = rgb[2] + mul255((short)p[0] - rgb[2], invalf[2]);
-    p[1] = rgb[1] + mul255((short)p[1] - rgb[1], invalf[1]);
-    p[2] = rgb[0] + mul255((short)p[2] - rgb[0], invalf[0]);
-#elif defined EFL_1BPP
+
+#ifdef EFL_1BPP
     int gray = grayscale( rgb[0], rgb[1], rgb[2] );
     int invalfgray = grayscale( invalf[0], invalf[1], invalf[2] );
     p[0] = gray + mul255((short)p[0] - gray, invalfgray);
@@ -468,7 +464,9 @@ void gli_draw_pixel_lcd(int x, int y, unsigned char *alpha, unsigned char *rgb)
     p[0] = rgb[2] + mul255((short)p[0] - rgb[2], invalf[2]);
     p[1] = rgb[1] + mul255((short)p[1] - rgb[1], invalf[1]);
     p[2] = rgb[0] + mul255((short)p[2] - rgb[0], invalf[0]);
+#ifndef WIN32
     p[3] = 0xFF;
+#endif
 #endif
 }
 
@@ -511,17 +509,15 @@ void gli_draw_clear(unsigned char *rgb)
         p = gli_image_rgb + y * gli_image_s;
         for (x = 0; x < gli_image_w; x++)
         {
-#ifdef WIN32
-            *p++ = rgb[2];
-            *p++ = rgb[1];
-            *p++ = rgb[0];
-#elif defined EFL_1BPP
+#ifdef EFL_1BPP
             *p++ = gray;
 #else
             *p++ = rgb[2];
             *p++ = rgb[1];
             *p++ = rgb[0];
+#ifndef WIN32
             *p++ = 0xFF;
+#endif
 #endif
         }
     }
@@ -554,17 +550,15 @@ void gli_draw_rect(int x0, int y0, int w, int h, unsigned char *rgb)
         unsigned char *p = p0;
         for (x = x0; x < x1; x++)
         {
-#ifdef WIN32
-            *p++ = rgb[2];
-            *p++ = rgb[1];
-            *p++ = rgb[0];
-#elif defined EFL_1BPP
+#ifdef EFL_1BPP
             *p++ = gray;
 #else
             *p++ = rgb[2];
             *p++ = rgb[1];
             *p++ = rgb[0];
+#ifndef WIN32
             *p++ = 0xFF;
+#endif
 #endif
         }
         p0 += gli_image_s;
@@ -923,17 +917,15 @@ void gli_draw_picture(picture_t *src, int x0, int y0, int dx0, int dy0, int dx1,
 #ifdef EFL_1BPP
             unsigned char sgray = grayscale(sr, sg, sb);
 #endif
-#ifdef WIN32
-            dp[x*3+0] = sb + mul255(dp[x*3+0], na);
-            dp[x*3+1] = sg + mul255(dp[x*3+1], na);
-            dp[x*3+2] = sr + mul255(dp[x*3+2], na);
-#elif defined EFL_1BPP
+#ifdef EFL_1BPP
             dp[x] = sgray + mul255(dp[x], na);
 #else
             dp[x*4+0] = sb + mul255(dp[x*4+0], na);
             dp[x*4+1] = sg + mul255(dp[x*4+1], na);
             dp[x*4+2] = sr + mul255(dp[x*4+2], na);
+#ifndef WIN32
             dp[x*4+3] = 0xFF;
+#endif
 #endif
         }
         sp += src->w * 4;
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -951,16 +951,17 @@ void gli_draw_picture(picture_t *src, int x0, int y0, int dx0, int dy0, int dx1,
             unsigned char sb = mul255(sp[x*4+2], sa);
 #ifdef EFL_1BPP
             unsigned char sgray = grayscale(sr, sg, sb);
-#endif
-#ifdef EFL_1BPP
+
             dp[x] = sgray + mul255(dp[x], na);
+#elifdef WIN32
+            dp[x*3+0] = sb + mul255(dp[x*3+0], na);
+            dp[x*3+1] = sg + mul255(dp[x*3+1], na);
+            dp[x*3+2] = sr + mul255(dp[x*3+2], na);
 #else
             dp[x*4+0] = sb + mul255(dp[x*4+0], na);
             dp[x*4+1] = sg + mul255(dp[x*4+1], na);
             dp[x*4+2] = sr + mul255(dp[x*4+2], na);
-#ifndef WIN32
             dp[x*4+3] = 0xFF;
-#endif
 #endif
         }
         sp += src->w * 4;
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -91,7 +91,7 @@ struct font_s
 
 static unsigned short gammamap[256];
 static unsigned char gammainv[256 << GAMMA_SHIFT];
-static unsigned char filterweights = {28, 56, 85, 56, 28};
+static unsigned char filterweights[5] = {28, 56, 85, 56, 28};
 
 static font_t gfont_table[8];
 
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -105,7 +105,7 @@ unsigned char *gli_image_rgb = NULL;
 
 #ifdef EFL_1BPP
 static const int gli_bpp = 1;
-#elifdef WIN32
+#elif defined WIN32
 static const int gli_bpp = 3;
 #else
 static const int gli_bpp = 4;
@@ -949,7 +949,7 @@ void gli_draw_picture(picture_t *src, int x0, int y0, int dx0, int dy0, int dx1,
             unsigned char sgray = grayscale(sr, sg, sb);
 
             dp[x] = sgray + mul255(dp[x], na);
-#elifdef WIN32
+#elif defined WIN32
             dp[x*3+0] = sb + mul255(dp[x*3+0], na);
             dp[x*3+1] = sg + mul255(dp[x*3+1], na);
             dp[x*3+2] = sr + mul255(dp[x*3+2], na);
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -36,7 +36,10 @@
 #include <math.h> /* for pow() */
 #include "uthash.h" /* for kerning cache */
 
+#define GAMMA_SHIFT 2
+
 #define mul255(a,b) (((a) * (b)) >> 8)
+#define mul1023(a,b) (((a) * (b)) >> 10)
 #define grayscale(r,g,b) ((30 * (r) + 59 * (g) + 11 * (b)) / 100)
 
 #ifdef _WIN32
@@ -86,8 +89,8 @@ struct font_s
  * Globals
  */
 
-static unsigned char gammamap[256];
-static unsigned char gammainv[256];
+static unsigned short gammamap[256];
+static unsigned char gammainv[256 << GAMMA_SHIFT];
 const unsigned char filterweights[5] = {28, 56, 85, 56, 28};
 
 static font_t gfont_table[8];
@@ -380,6 +383,7 @@ static void loadfont(font_t *f, char *name, float size, float aspect, int style)
 
 void gli_initialize_fonts(void)
 {
+    const int gammamax = (256 << GAMMA_SHIFT) - 1;
     float monoaspect = gli_conf_monoaspect;
     float propaspect = gli_conf_propaspect;
     float monosize = gli_conf_monosize;
@@ -388,10 +392,10 @@ void gli_initialize_fonts(void)
     int i;
 
     for (i = 0; i < 256; i++)
-        gammamap[i] = (unsigned char)(pow(i / 255.0, gli_conf_gamma) * 255.0 + 0.5);
+        gammamap[i] = (unsigned short)(pow(i / 255.0, gli_conf_gamma) * gammamax);
 
-    for (i = 0; i < 256; i++)
-        gammainv[i] = (unsigned char)(pow(i / 255.0, 1.0 / gli_conf_gamma) * 255.0 + 0.5);
+    for (i = 0; i <= gammamax; i++)
+        gammainv[i] = (unsigned char)(pow(i / (float)gammamax, 1.0 / gli_conf_gamma) * 255.0);
 
     err = FT_Init_FreeType(&ftlib);
     if (err)
@@ -431,14 +435,15 @@ void gli_initialize_fonts(void)
 
 void gli_draw_pixel(int x, int y, unsigned char alpha, unsigned char *rgb)
 {
+    const int gammamax = (256 << GAMMA_SHIFT) - 1;
     unsigned char *p = gli_image_rgb + y * gli_image_s + x * gli_bpp;
-    unsigned char invalf = 255 - alpha;
-    unsigned char bg[3] = {
+    unsigned short invalf = gammamax - (alpha << GAMMA_SHIFT);
+    unsigned short bg[3] = {
         gammamap[p[0]],
         gammamap[p[1]],
         gammamap[p[2]]
     };
-    unsigned char fg[3] = {
+    unsigned short fg[3] = {
         gammamap[rgb[0]],
         gammamap[rgb[1]],
         gammamap[rgb[2]]
@@ -451,11 +456,11 @@ void gli_draw_pixel(int x, int y, unsigned char alpha, unsigned char *rgb)
 
 #ifdef EFL_1BPP
     int gray = grayscale( fg[0], fg[1], fg[2] );
-    p[0] = gammainv[gray + mul255((short)bg[0] - gray, invalf)];
+    p[0] = gammainv[gray + mul1023(bg[0] - gray, invalf)];
 #else
-    p[0] = gammainv[fg[2] + mul255((short)bg[0] - fg[2], invalf)];
-    p[1] = gammainv[fg[1] + mul255((short)bg[1] - fg[1], invalf)];
-    p[2] = gammainv[fg[0] + mul255((short)bg[2] - fg[0], invalf)];
+    p[0] = gammainv[fg[2] + mul1023(bg[0] - fg[2], invalf)];
+    p[1] = gammainv[fg[1] + mul1023(bg[1] - fg[1], invalf)];
+    p[2] = gammainv[fg[0] + mul1023(bg[2] - fg[0], invalf)];
 #ifndef WIN32
     p[3] = 0xFF;
 #endif
@@ -464,17 +469,18 @@ void gli_draw_pixel(int x, int y, unsigned char alpha, unsigned char *rgb)
 
 void gli_draw_pixel_lcd(int x, int y, unsigned char *alpha, unsigned char *rgb)
 {
+    const int gammamax = (256 << GAMMA_SHIFT) - 1;
     unsigned char *p = gli_image_rgb + y * gli_image_s + x * gli_bpp;
-    unsigned char invalf[3];
-        invalf[0] = 255 - alpha[0];
-        invalf[1] = 255 - alpha[1];
-        invalf[2] = 255 - alpha[2];
-    unsigned char bg[3] = {
+    unsigned short invalf[3];
+        invalf[0] = gammamax - (alpha[0] << GAMMA_SHIFT);
+        invalf[1] = gammamax - (alpha[1] << GAMMA_SHIFT);
+        invalf[2] = gammamax - (alpha[2] << GAMMA_SHIFT);
+    unsigned short bg[3] = {
         gammamap[p[0]],
         gammamap[p[1]],
         gammamap[p[2]]
     };
-    unsigned char fg[3] = {
+    unsigned short fg[3] = {
         gammamap[rgb[0]],
         gammamap[rgb[1]],
         gammamap[rgb[2]]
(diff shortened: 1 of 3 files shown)


Tynach:
Dark text on a light background will appear to be more pale than before (when using a gamma value higher than 1.0), with less contrast especially on the dark blue links produced by the default configuration file. However, this is a side effect caused by the text being rendered more correctly.

There also shouldn't be any perceptible difference when using a gamma of 1.0, which is the default setting. Thus, anyone who never touched that setting shouldn't notice anything change.

Tynach:
After having slept since implementing this, I've had a few thoughts. Namely, I'm curious how open you guys would be to adding some new configuration options for how FreeType should render the text. Here are some ideas:

1. Perhaps the filter weights can be a setting instead of the admittedly kludgy 'is gamma equal to 1?' check I have here.
    1. Come to think of it, even if the 'kludge' is kept, I wonder if the check should be to make sure gamma is *over* 1 before switching to the `light` LCD filter. Or perhaps check if it's over, say, 1.5.

Tynach:
After having slept since implementing this, I've had a few thoughts. Namely, I'm curious how open you guys would be to adding some new configuration options for how FreeType should render the text. Here are some ideas:

1. Perhaps the filter weights can be a setting instead of the admittedly kludgy 'is gamma equal to 1?' check I have here.

    1. Come to think of it, even if the 'kludge' is kept, I wonder if the check should be to make sure gamma is *over* 1 before switching to the `light` LCD filter. Or perhaps check if it's over, say, 1.5.

2. I have my own monitors calibrated to use sRGB's transfer curves, and would love to have the option to use that as well. However:

    1. 'gamma' is, I presume, a numeric field only. I haven't looked at how settings are retrieved from the .ini file though, and don't know what happens when I try putting a text string there.. Except that it breaks rendering spectacularly.

    2. If possible, I'd like to give several textual options. Among them:
        1. sRGB
        2. Rec. 709/SMPTE 170M (probably just check for the text `rec709`)
        3. Maayybe L\*a\*b\* (popular but outdated approximation for human perceptual vision; not used by any monitors to my knowledge, but has a mathematical formula similar to sRGB's)

    As tempting as it is to also add BT.1886, that would require *even more* parameters, as BT.1886's curve changes depending on the contrast ratio of the display in question. Or rather, as inputs it requires the candelas per square meter (AKA cd/m^2, or 'nits') at the surface of the display, both when the display is showing black and when the display is showing white.

    Those parameters *could* be obtained from an ICC profile if one is present, but if we do that, then...

3. Option to obtain gamma and inverse gamma lookup tables from ICC profile.

    When a display is profiled and an ICC/ICM file is generated, to my knowledge the resulting ICC file almost always includes the relevant lookup tables directly. I don't have a *ton* of experience with this, admittedly, as I mostly just use DisplayCal for profiling my monitors and I don't know how much it spoils me compared to other applications.

    However, the downside to this is that we'd have to deal with multiple monitors (I have 2), and potentially have to swap which gamma lookup tables are being used if/when the user moves Gargoyle from one monitor to another. Another option is to just use the monitor that Gargoyle starts up on, or add yet another configuration option to Gargoyle to allow the user to select which monitor's display profile to pull the information from.

4. In fact, it might just be easier to let the user point to the path of an actual ICC profile to use. ICC profiles could even provide gamma tables for all the various options in item 2 of this list, making it unnecessary to specifically support those things.

    However, the gamma tables in ICC profiles are meant to also be used with other colorimetry data in those profiles.. And there is a *lot* of data that can be included in those files. The ICC profile I use for my primary monitor is 3.6 MiB in size (though other times I've profiled it I've gotten various other sizes, depending on my settings and the accuracy I wanted; another profile for the same display is just 486.1 MiB, and others just a hundred or so KiB).

    In general, ICC profiles are a bit overkill for this, though they're certainly a viable option.

Tynach:
I was being rushed to go somewhere when typing that comment, and also forgot to mention some other options I'd like to see considered for implementation:

5. Option to enable 'Slight Hinting'

    I know, I know... One of the whole points behind Gargoyle's manual FreeType handling is that fonts are as close as possible to their original shapes and weights.

    However, especially with proper gamma correct rendering, thin fonts (such as many serif fonts) are often barely visible and hard to read. Additionally, FreeType in particular has implemented many features that help with this while also retaining the shapes of fonts as much as possible.

    In particular, FreeType now offers 'slight' hinting, which only hints fonts vertically. This allows for the letter spacing between each glyph to remain unchanged, as well as much of the general shape of the characters. However, what *will* get shifted are things like the top of a 'T' or the middle bar of an 'H' to help make sure they don't land between pixels, which leads to unnecessarily blurry horizontal lines.

6. Option to enable 'stem darkening'

    This one is best explained through examples. Thankfully, FreeType's developers wrote [a fairly lengthy article on the topic][sd], which also covers 'slight hinting' as mentioned above (or more specifically, how it changed in FreeType 2.6.2). The article includes a number of images, though I want to point out [this one in particular][sdpic] (best viewed in its own tab at 100% (1:1) zoom so that pixels don't get incorrectly interpolated by your browser).

    Sadly, there are a few caveats to this at the moment.. Namely, that it doesn't yet work with TrueType fonts without using the autohinter. Still, I think there's merit to providing the option to enable it, as well as the autohinter in at least 'slight hinting' mode so that it can be used with any font. I believe that's what's being done in that image.

The reason I bring these up is because some people might consider the lighter text while using gamma correction to be a bug, but we could use these features of FreeType to help mitigate the issue.
___

Side note: the comparison image I linked to also happens to be using grayscale rendering, which I also happened to test in Gargoyle while writing those commits. Grayscale antialiasing in gargoyle appears to have always been a bit more accurate (though only in the black-text-on-white-background scenario) than LCD antialiasing, but also much *uglier* for some reason. Even with gamma correct rendering, black text on white backgrounds had jagged lines and few levels of gray being interpolated.

This is fixed kinda by accident in my version of the code. I say 'kinda' because I did purposefully apply my changes to all the rendering paths I saw, but I only actually tested it and noticed how much improvement was made *after* I'd committed the changes. It was an afterthought to check the code before/after while using grayscale antialiasing, rather than something I intentionally set out to fix.

[sd]: https://www.freetype.org/freetype2/docs/hinting/text-rendering-general.html
[sdpic]: https://www.freetype.org/image/BlendingExamples.png

Tynach:
Okay, I see that the error it fails at is:

```
garglk/draw.c:94:7: error: unknown type name 'FT_LcdFiveTapFilter'
 const FT_LcdFiveTapFilter filterweights = {28, 56, 85, 56, 28};
       ^
```

`FT_LcdFiveTapFilter` is defined in `freetype/ftlcdfil.h`, which I included on line 34 (via the preprocessor-defined constant `FT_LCD_FILTER_H`).

I don't know why it's not detecting that in the automated build. I also am not very familiar with Jam in general, and don't feel qualified to poke at it.

Tynach:
I'm going over my math.. My version of the gamma correction code, as it stands, is still not completely accurate at low values. At a gamma of 2.2, the last inaccurate value is 20; at a gamma of 1.8, the last one is 6. So only very low values are affected, but it's enough to make me feel I should mention it.

I don't know if you guys consider color accuracy of images as more or less important than the text rendering. I just thought I'd mention my findings.

garglk/draw.c
```diff
@@ -212,10 +211,18 @@ static void loadglyph(font_t *f, glui32 cid)
         if (f->make_oblique)
             FT_Outline_Transform(&f->face->glyph->outline, &ftmat);
 
-        if (gli_conf_lcd)
+        if (gli_conf_lcd) {
+            if (gli_conf_gamma == 1.0) {
+                FT_Library_SetLcdFilterWeights(ftlib, filterweights);
```
cspiegel, line 216: This results in a diagnostic because `FT_LibrarySetLcdFilterWeights()` expects `unsigned char *` but this is passing the equivalent of `const unsigned char*`.  Clearly a bug in the API, but `filterweights` should have the const removed to appease it.

`filterweights` also only used in one place, so it could be moved directly into this block instead of being global, but if you prefer keeping it global, please add `static` to it.

cspiegel:
I ran a test with light on dark and dark on light, and in both cases the rendering was subtly, but obviously, improved (it looks sharper in both cases).  This is fantastic, thank you!

It works great for me on Linux with both EFL and gtk as the rendering engine.  However, it looks garbled on Windows:

windows

I'm looking at it but you probably would be able to hone in on it more quickly.

Tynach:
@cspiegel, uh.. I have no clue what's causing that. I've not booted Windows for quite some time either, so it might not be very fast to have me try debugging it.

That said, I have a suspicion. The first commit of the pull request rearranges some preprocessor directives to reduce what I thought was needlessly duplicated code. I did that because I knew I'd be modifying a few of those parts, and changed the rest in order to keep them all consistent.

Try checking out commit 5259a78fb566077c3f090a771c34dc50dee57aa8 and see if the garbled output remains. If so, I messed that up and should probably have stuck to just doing the minimum required changes to implement the new gamma code.

As a further test, however.. If it is still garbled with only that commit, try the commit before it (98dae6b95fa14ac439be9f1ff9aa8d412ac16538) and see if the problem persists. Note that my pull request doesn't include the MinGW changes added in 0bac2f416748ee3109ab392610f1efbcdeb8b1ee.

I remember that some Electron apps (notably Discord) had garbled text rendering on Linux when using LCD antialiasing for a while, because they were built against one version of FreeType but many Linux distros used a different version of FreeType. Basically, make sure the FreeType headers match the version of the FreeType library you're using on that Windows build.
___

By the way, I might talk a lot about this type of thing, but honestly that's because I was basically mildly obsessed with colorspace conversion code for several years straight out of boredom, since I've been unemployed. So, while I know quite a bit about gamma correct rendering (and both how it works and should look), this is honestly the first time I've ever attempted to implement it using a lookup table - and the first time I've tried implementing it for an actual existing application to use.

It's also the first time I've touched the FreeType API directly; been meaning to mess around and learn how to use it for a while now, so this has been quite fun! I just really hope I don't mess things up *too* much for you guys.

Tynach:
By the way, what are the circumstances in which `EFL_1BPP` is defined? It might be worth testing builds in that category as well. It looks almost as if things in that screenshot are being rendered with 4 components per pixel, which indeed is what a lot of those preprocessor directives I rearranged deal with. According to them, only Windows should have 3 components, and everyone else should have 4 except for `EFL_1BPP`, which should have 1 (grayscale).

If this is the problem, I wonder if such a build shows even worse (though similar) behavior.

Tynach:
Ah, I think I see the problem. The multiplication math is different in the two groups in the last one I changed in that commit. `WIN32` should have things multiplied by 3, and everyone else should have things multiplied by 4. The goof is at line 920.

I'll change that part back to how it was originally.

Tynach:
@cspiegel I think I've fixed it. Ended up not restoring it *exactly* as it was; felt silly to not combine the two parts with `EFL_1BPP` when they could logically be placed right next to each other. Still, I think it'd be worth testing such a build just to make sure. I can at least confirm that Linux builds still work.

Tynach:
By the way @cspiegel, you said there was only a subtle difference in rendering in both cases on Linux? Did you make sure to have `gamma` in `garglkrc`/`garglk.ini` set to something higher than 1.0? Or would you consider the difference between the first two images in https://github.com/garglk/garglk/pull/419#issuecomment-839210320 to also be subtle?

I can understand that being subtle to some people; it isn't to me personally, but that's probably because it's something I've studied for a while and know what to look for (and might subconsciously look for). My sense of what is and isn't subtle might be kinda warped.

cspiegel:
I'll try to answer all your questions/issues.  If I've missed something, please let me know.

Your Windows fix is correct, although there was another problem causing the overall issues I saw.  `#elifdef` is not supported in C (it will be in the next revision, but isn't currently), so that was causing issues.  The use of `#elifdef` needs to be `#else if defined`.  With that fix, most things looked right; all except picture drawing.  But your commit fixed picture drawing, so the Windows side looks good now.  Have you, by chance, been able to test on macOS?  I don't have access to a development environment on macOS so I can't test, but on the other hand there's no special case for it, so if Linux works, I'd expect macOS to as well.

`EFL_1BPP` is defined when the target OS is Open Inkpot.  That was a Linux distribution for use with e-ink readers.  Apparently it used EFL as the UI library.  24-bit support was added to the EFL backend as well, likely because it was so trivial to do so, for anybody wanting to use EFL on the desktop.  Open Inkpot is dead and I've thought about removing EFL support, but since I can still test it (the 24-bit, not 1), I've kept it around.  At the very least removing 1bpp support is probably a good idea, though I'm really leaning toward just ripping EFL out entirely.

You're right about the gamma when I was testing: I have mine set at 1.0.  When I changed parameters to be similar to yours, then the output looks a lot more in line with your examples.  Sorry for the noise there, but it's still nice to see that at 1.0, the text is a bit crisper.

Adding new configuration options is no problem at all.  We'll have to keep "gamma" as-is, at least when its argument is numeric, for backward compatibility.  But I'm all for adding specific knobs for tuning fonts.

I'm certainly not against adding ICC profile support, if it has a tangible benefit: I just looked up what they are right now, so I'm not qualified to speak to their utility, but I'm not against giving more options for producing better output.  They can simply be ignored by anybody who doesn't have interest in them.

Fontconfig is only used on Unix, so it's probably better to use it solely for finding font files, and keep all rendering options in the Gargoyle config file.  Much as I like being able to set font options in `~/.config/fontconfig/conf.d`, Gargoyle is a bit of a special case, in that it's not unlikely that you'll want it to be configured differently than your system.  But really, the fact that we're cross-platform does mean that the best location for settings is Gargoyle's own config.

Slight hintingh/stem darkening: as above, new config options are completely fine.  I do think stem darkening would be a welcome option: I've wrestled with problems due to a lack of stem darkening in most places (the exception being Qt, which does stem darkening at a gamma of 1.8, resulting in better-looking OTF rendering, to my eye), so I'm in favor of allowing it to be enabled.

As for color accuracy of images vs text rendering, does one affect the other?  I suppose it all depends on what the tradeoff is.

garglk/draw.c
Tynach: I defined it globally because the original equivalent was `#define`d globally. Aand I forgot to remove them.. But they're up at lines 137-139 (originally; now lines 143-145).

GCC doesn't bother me about `const` on my end, but I can definitely change it to be `static` instead of `const` so the tests pass.

Tynach:
> I'll try to answer all your questions/issues. If I've missed something, please let me know.
> 
> Your Windows fix is correct, although there was another problem causing the overall issues I saw. `#elifdef` is not supported in C (it will be in the next revision, but isn't currently), so that was causing issues. The use of `#elifdef` needs to be `#elif defined`. With that fix, most things looked right; all except picture drawing. But your commit fixed picture drawing, so the Windows side looks good now. Have you, by chance, been able to test on macOS? I don't have access to a development environment on macOS so I can't test, but on the other hand there's no special case for it, so if Linux works, I'd expect macOS to as well.

I have no Apple hardware with which to test, sorry. That isn't likely to change either, given a continued lack of income.

> `EFL_1BPP` is defined when the target OS is Open Inkpot. That was a Linux distribution for use with e-ink readers. Apparently it used EFL as the UI library. 24-bit support was added to the EFL backend as well, likely because it was so trivial to do so, for anybody wanting to use EFL on the desktop. Open Inkpot is dead and I've thought about removing EFL support, but since I can still test it (the 24-bit, not 1), I've kept it around. At the very least removing 1bpp support is probably a good idea, though I'm really leaning toward just ripping EFL out entirely.

Huh, I see. If it was still something worth supporting, I was going to ask if it was 1 byte per pixel, or one bit per pixel; and in the case of one byte per pixel, I was going to ask if I should see about tweaking the grayscale multipliers to be more accurate; that's one thing that would get an accuracy boost when done with gamma correction, since you can use the chromaticity coordinates for the original RGB color space's primary colors to calculate exactly how much each of the three channels contributes to the luminosity of the color (which then translates to the desired gray level).

For sRGB, on a scale of 0 to 1, the coefficients would be roughly 0.212639 for red, 0.715169 for green, and 0.072192 for blue. The scale in the `grayscale(r,g,b)` macro seems to be 100, so the correct coefficients would roughly be `21, 72, 7`. If dividing by 255 (or, for use with the inverse gamma lookup table, 2047) at the end instead of 100, the values would be `54*r + 182*g + 18*b` (or `435*r + 1464*g + 148*b` for 2047).

> You're right about the gamma when I was testing: I have mine set at 1.0. When I changed parameters to be similar to yours, then the output looks a lot more in line with your examples. Sorry for the noise there, but it's still nice to see that at 1.0, the text is a bit crisper.

Huh. I wonder why; when gamma is left at 1, there shouldn't be any difference in how it's rendered.. Unless you also are affected by the bug caused by using FreeType 2.10.3 or later, and just didn't realize it. Looking back, I should have had a fix for that as the first commit, instead of fixing it along with the gamma correction in general.

> Adding new configuration options is no problem at all. We'll have to keep "gamma" as-is, at least when its argument is numeric, for backward compatibility. But I'm all for adding specific knobs for tuning fonts.

Well, adding support for things like sRGB and Rec. 709 would kinda 'override' the gamma setting. This is because they tend to be power curves with a linear segment at the start, and the curved part is 'squished' horizontally. That's why, for example, sRGB uses a power of 2.4 but is most closely approximated using a power of 2.2.

Perhaps we could do what the FreeType demo code used to do, and use `0.0` to refer to sRGB.

> I'm certainly not against adding ICC profile support, if it has a tangible benefit: I just looked up what they are right now, so I'm not qualified to speak to their utility, but I'm not against giving more options for producing better output. They can simply be ignored by anybody who doesn't have interest in them.

ICC support would almost certainly be overkill. I'd say it's unrealistic *if* we were to just implement it ourselves; if we instead implement it using something like LCMS, it'd be.. *Doable*, but likely require we rearrange how the rendering pipeline works. It'd still certainly be overkill for *only* text rendering, but if we were to color correct not just text, but also images drawn and whatnot, it might no longer qualify as 'overkill'.

Actually, one thing using LCMS *would* let us do is render everything as if it were being printed on paper. We'd just perform the transformations necessary for 'soft proofing' the result of translating the colors for use with a printer profile. That might be desirable for some people, though kinda niche.. And the only advantage to doing that with LCMS vs. rolling our own little effect filter, is that LCMS would let us do it with a high degree of accuracy.

> Fontconfig is only used on Unix, so it's probably better to use it solely for finding font files, and keep all rendering options in the Gargoyle config file. Much as I like being able to set font options in `~/.config/fontconfig/conf.d`, Gargoyle is a bit of a special case, in that it's not unlikely that you'll want it to be configured differently than your system. But really, the fact that we're cross-platform does mean that the best location for settings is Gargoyle's own config.

Bah. I'd want the rest of my system to look how Gargoyle does with these changes.. Stupid users made that unlikely for the near future, though.

> Slight hintingh/stem darkening: as above, new config options are completely fine. I do think stem darkening would be a welcome option: I've wrestled with problems due to a lack of stem darkening in most places (the exception being Qt, which does stem darkening at a gamma of 1.8, resulting in better-looking OTF rendering, to my eye), so I'm in favor of allowing it to be enabled.

Are you sure Qt does that still? I remember they made a change to do that, but a bunch of users complained about the lighter text, especially since stem darkening is only available to OTF fonts. So they pulled support for it from the Linux builds to appease the masses.

Maybe they enabled it again but only in QML or something like that. As far as I know, however, QWidgets applications still have a gamma of 1.0 and that can't be changed at runtime (to my knowledge). If you know of a way to change this PLEASE tell me how. This has been bugging me ever since I had found that 'bug' report.

I do know they still do it on Windows, but that's because Windows users are used to seeing a bit of gamma correction applied to text. Same for Mac OS users.

> As for color accuracy of images vs text rendering, does one affect the other? I suppose it all depends on what the tradeoff is.

Oh. This is because I implement the gamma correction itself inside of `gli_draw_pixel()` and `gli_draw_pixel_lcd()`, which are used for both drawing text *and* images. I did this because it's where the alpha blending occurs.. In fact, the functions don't exactly draw a pixel at a set value, but instead composite one pixel on top of another existing one, using an alpha value to blend between the two colors (and that alpha value is also a color in the `lcd` variant).

I basically implemented linear alpha blending for everything that uses those functions, so it does affect the drawing of images as well as text. At first I thought this would affect PNG images with transparency being drawn, but I haven't observed any changes in the few games I've tried that have PNG images drawn.

I haven't looked at the PNG handling code, but it seems as if it's calculating the alpha blending first before sending the result off to actually be drawn. Is this accurate?

cspiegel:
> Huh, I see. If it was still something worth supporting, I was going to ask if it was 1 byte per pixel, or one bit per pixel; and in the case of one byte per pixel, I was going to ask if I should see about tweaking the grayscale multipliers to be more accurate; that's one thing that would get an accuracy boost when done with gamma correction, since you can use the chromaticity coordinates for the original RGB color space's primary colors to calculate exactly how much each of the three channels contributes to the luminosity of the color (which then translates to the desired gray level).

I doubt it's worth keeping around.  Probably not worth the effort to fix up grayscale handling.  Open Inkpot is a dead project, so there's not much point in trying to get it working.  In fact, I think a lot of updates to EFL have been done blindly, just because it's so unused.  This is really an impetus for me to just remove it.

> Huh. I wonder why; when gamma is left at 1, there shouldn't be any difference in how it's rendered.. Unless you also are affected by the bug caused by using FreeType 2.10.3 or later, and just didn't realize it. Looking back, I should have had a fix for that as the first commit, instead of fixing it along with the gamma correction in general.

I'm using Freetype 2.10.4.  The original code produces the following:

original-fs8

The new code looks like this:

new-fs8

But as you say, that could just be the Freetype fixup.  Either way, it looks better, which is nice.

> Well, adding support for things like sRGB and Rec. 709 would kinda 'override' the gamma setting. This is because they tend to be power curves with a linear segment at the start, and the curved part is 'squished' horizontally. That's why, for example, sRGB uses a power of 2.4 but is most closely approximated using a power of 2.2.

It'd be fine that a new setting overrides an existing one, just as long as the default stays as-is.  It's OK to say "if you set this, gamma will be ignored", for example.

> ICC support would almost certainly be overkill. I'd say it's unrealistic if we were to just implement it ourselves; if we instead implement it using something like LCMS, it'd be.. Doable, but likely require we rearrange how the rendering pipeline works. It'd still certainly be overkill for only text rendering, but if we were to color correct not just text, but also images drawn and whatnot, it might no longer qualify as 'overkill'.

This is beyond my area of knowledge.  In general, if tangible benefits are there and somebody's willing to do the work, then it can be incorporated.  Up to you to determine if it's worth it!

> Are you sure Qt does that still? I remember they made a change to do that, but a bunch of users complained about the lighter text, especially since stem darkening is only available to OTF fonts. So they pulled support for it from the Linux builds to appease the masses.

The font situation on Linux is a mess.  I went through this long ago when I realized Konsole's rendering of OTF fonts looked better, to my eye, than other terminal emulators (such as Alacritty).  I dug through the Qt source, but forgot where things were, so I just did it again (this is on the `kde/5.15` branch, but it's been this way for a while):

`fontSmoothingGamma` is set to the result of: `QGuiApplicationPrivate::platformIntegration()->styleHint(QPlatformIntegration::FontSmoothingGamma).toReal()`

So gamma correction is done on a per-platform basis.  `QXcbIntegration::styleHint` does _not_ handle `QPlatformIntegration::FontSmoothingGamma`, instead delegating to `QPlatformIntegration::styleHint()`.  _That_ function (in `src/gui/kernel/qplatformintegration.cpp`) returns `1.7` for the gamma (so I slightly misremembered: they use 1.7 instead of 1.8).

`src/platformsupport/fontdatabases/freetype/qfontengine_ft.cpp` also does this:

```
FT_Bool no_darkening = false;
FT_Property_Set(freetypeData->library, "cff", "no-stem-darkening", &no_darkening);
```
This was done as part of commit 21f27c9.  So Qt did, at one point, disable gamma correction, but brought it back with stem darkening.

> I basically implemented linear alpha blending for everything that uses those functions, so it does affect the drawing of images as well as text. At first I thought this would affect PNG images with transparency being drawn, but I haven't observed any changes in the few games I've tried that have PNG images drawn.

> I haven't looked at the PNG handling code, but it seems as if it's calculating the alpha blending first before sending the result off to actually be drawn. Is this accurate?

I don't know (I'm not the author of the code, nor have I ever looked at it!)  Basically, you're in territory that I've never touched, so unfortunately I can't offer any insight there.

cspiegel:
@angstsmurf Would you be willing to give this a quick build and test on macOS to ensure the drawing code is fine there?

Tynach:
> I'm using Freetype 2.10.4. The original code produces the following:
> 
> original-fs8
> 
> The new code looks like this:
> 
> new-fs8
> 
> But as you say, that could just be the Freetype fixup. Either way, it looks better, which is nice.

Yep, that's the issue. FreeType 2.10.4 has the new behavior, so the text was getting filtered twice (once by FreeType, once by Gargoyle).

Here's the output of `git diff` if I checkout the last commit before my changes, and then only fix that bug (so you can see how Gargoyle rendered text before FreeType changed its API):

<details>
<summary>Click to expand <code>git diff</code> output</summary>

```diff
diff --git a/garglk/draw.c b/garglk/draw.c
index 82ff092..cdc51fa 100644
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -31,6 +31,7 @@
 #include <ft2build.h>
 #include FT_FREETYPE_H
 #include FT_OUTLINE_H
+#include FT_LCD_FILTER_H
 
 #include <math.h> /* for pow() */
 #include "uthash.h" /* for kerning cache */
@@ -212,10 +213,12 @@ static void loadglyph(font_t *f, glui32 cid)
         if (f->make_oblique)
             FT_Outline_Transform(&f->face->glyph->outline, &ftmat);
 
-        if (gli_conf_lcd)
+        if (gli_conf_lcd) {
+            FT_Library_SetLcdFilter(ftlib, FT_LCD_FILTER_NONE);
             err = FT_Render_Glyph(f->face->glyph, FT_RENDER_MODE_LCD);
-        else
+        } else
             err = FT_Render_Glyph(f->face->glyph, FT_RENDER_MODE_LIGHT);
+
         if (err)
             winabort("FT_Render_Glyph");
 

```

</details>

> It'd be fine that a new setting overrides an existing one, just as long as the default stays as-is. It's OK to say "if you set this, gamma will be ignored", for example.

Noted!

> This is beyond my area of knowledge. In general, if tangible benefits are there and somebody's willing to do the work, then it can be incorporated. Up to you to determine if it's worth it!

For a text adventure game front-end? I don't think it is. A small part of my brain is yelling at me because it *would* be super cool and all, but unless someone else does it (or helps me do it) and it's shown that accurately correcting colors from sRGB to arbitrary display devices or simulating printer ink is desirable by more people than just a weirdo like me, I don't think it'd be worth it.

> The font situation on Linux is a mess. I went through this long ago when I realized Konsole's rendering of OTF fonts looked better, to my eye, than other terminal emulators (such as Alacritty). I dug through the Qt source, but forgot where things were, so I just did it again (this is on the `kde/5.15` branch, but it's been this way for a while):
> 
> `fontSmoothingGamma` is set to the result of: `QGuiApplicationPrivate::platformIntegration()->styleHint(QPlatformIntegration::FontSmoothingGamma).toReal()`
> 
> So gamma correction is done on a per-platform basis. `QXcbIntegration::styleHint` does _not_ handle `QPlatformIntegration::FontSmoothingGamma`, instead delegating to `QPlatformIntegration::styleHint()`. _That_ function (in `src/gui/kernel/qplatformintegration.cpp`) returns `1.7` for the gamma (so I slightly misremembered: they use 1.7 instead of 1.8).
> 
> `src/platformsupport/fontdatabases/freetype/qfontengine_ft.cpp` also does this:
> 
> ```
> FT_Bool no_darkening = false;
> FT_Property_Set(freetypeData->library, "cff", "no-stem-darkening", &no_darkening);
> ```
> 
> This was done as part of commit 21f27c9. So Qt did, at one point, disable gamma correction, but brought it back with stem darkening.

I looked into that commit. Seems they're *only* performing gamma correction when stem darkening is available, specifically only when the CFF engine is used (OTF and Type1 fonts). I'd still like to see this enabled for TTF fonts - even *without* stem darkening.. But it's fantastic that it's finally available at all.

Also had to go test it out, and fonts *finally* are indeed rendered with gamma correction, at least under certain conditions! I do use KDE, so I might be able to start replacing fonts with OTF versions to get gamma correction in most apps that I use regularly.

> I don't know (I'm not the author of the code, nor have I ever looked at it!) Basically, you're in territory that I've never touched, so unfortunately I can't offer any insight there.

I can see toward the bottom of `gli_draw_picture()` that it calculates the alpha value using the fourth color channel.. But it's possible there's some processing going on before it gets to that point. I'm really unsure.

Is there a game you can recommend I test this with? Something that draws either two PNGs on top of each other (the upper one having transparency), or draws a png with transparency on top of text, would be ideal.

I'll be honest, shortly before filing bug #418 was basically my first forray into interpreted interactive fiction games. I've played some interactive fiction games before, but usually they were either flash games or actual binaries - not something I'd run in a third party application. I've thus only been exposed to a few of them, several of which are NSFW and seemed inappropriate for screenshots (had to do some research and find other games before actually posting any).

As a result, I'm not entirely sure what sorts of things are possible in this format, or how many of these games 'should' render. Screenshots seem varied, and there are surprisingly few around for the more graphically heavy IF games I *have* found.

angstsmurf:
> @angstsmurf Would you be willing to give this a quick build and test on macOS to ensure the drawing code is fine there?

It seems to be working!

There are some problems with the macOS build script on the lastest version. I'll open a separate issue for them.

cspiegel:
> Is there a game you can recommend I test this with? Something that draws either two PNGs on top of each other (the upper one having transparency), or draws a png with transparency on top of text, would be ideal.

I don't know if any games do that, but there is a more flexible approach that can be used instead (if a bit more cumbersome).  You can write a Glk program in C that does exactly what you need.  It's a bit involved, but not horribly so.

First, get blorbtool from https://eblong.com/zarf/blorb/.  Run it something like this:

```
python3 blorbtool.py -n -o out.blb
> import Pict 1 PNG /path/to/file.png
> import Pict 2 PNG /path/to/another.png
> save
```
This will create a blorb file called `out.blb` which contains two pictures with numbers 1 and 2.  You can add as many as you want.  Then the following C program:

```c
#include <assert.h>  
  
#include <glk.h>  
#include <glkstart.h>  
#include <gi_blorb.h>  
  
void glk_main(void)  
{  
    strid_t file = glkunix_stream_open_pathname("out.blb", 0, 0);  
    assert(file != NULL);  
    assert(giblorb_set_resource_map(file) == giblorb_err_None);  
  
    winid_t win = glk_window_open(0, 0, 0, wintype_Graphics, 0);  
    assert(win != NULL);  
  
    glk_image_draw(win, 1, 0, 0);  
    glk_image_draw(win, 2, 5, 5);  
}  
  
int glkunix_startup_code(glkunix_startup_t *data)  
{  
    return 1;  
}
```
Build with something like:

```
cc -Wall -O glk.c -I$HOME/github/garglk/garglk -L$HOME/github/garglk/build/linux.release/garglk -lgarglk -lgarglkmain
```
Adjusting the path to Gargoyle.

The program draws picture 1 at 0x0 and picture 2 at 5x5, overlapping them.  This will allow you to use arbitrary PNGs and combine them however you'd like.

As for images on text, I'm not even sure that Glk supports that.

cspiegel:
> It seems to be working!

Great, thanks for testing!

> There are some problems with the macOS build script on the lastest version, however. I'll open a separate issue for them.

Sounds good.  One of these days I'll get access to a Mac so I can ensure things work across all platforms.

Tynach:
> The program draws picture 1 at 0x0 and picture 2 at 5x5, overlapping them. This will allow you to use arbitrary PNGs and combine them however you'd like.

Huh, interesting. I'll give it a go once I've come up with a couple of png files to use.

> As for images on text, I'm not even sure that Glk supports that.

I *think* it's supposed to support text on top of an image at least, though Gargoyle doesn't seem to work with that. At the very least, there's a game called [Narcolepsy] that I've found [a screenshot of][narco-1] showing a thought bubble backdrop for the text on the right side. Gargoyle doesn't draw that thought bubble backdrop, though - not even using a version of the code from before my changes.

And just now I found [another screenshot][narco-2] of that game that shows a further backdrop, which.. Looks like the wallpaper of Windows Vista? Is it perhaps trying to render that part of the window transparent?

... I'm gonna try setting Gargoyle to have a window background color of black and see what happens.

[Narcolepsy]: https://ifdb.org/viewgame?id=b67iuep7o6b0kxpm
[narco-1]: http://www.davidkinder.co.uk/glulxe_narco.png
[narco-2]: https://emshort.files.wordpress.com/2017/05/glulxe_narco.png

angstsmurf:
> there's a game called [Narcolepsy](https://ifdb.org/viewgame?id=b67iuep7o6b0kxpm) that I've found [a screenshot of](http://www.davidkinder.co.uk/glulxe_narco.png) showing a thought bubble backdrop for the text on the right side.

I'm pretty sure that is just four border images around a square text-only window with a white background.

Tynach:
> I'm pretty sure that is just four border images around a square text-only window with a white background.

Aah, ok. Strange it doesn't render.

angstsmurf:
You should also be able to prevent the game from overriding background color and other text attributes by setting `stylehint 0`in the `garglk.ini` file.

Tynach:
> You should also be able to prevent the game from overriding background color and other text attributes by setting `stylehint 0`in the `garglk.ini` file.

That prevents it from having gray text on a blue background on the left side, but doesn't let the border images on the right side show. The top 'logo' image on the left side always shows, though, so there's that.

Either way, kinda outside the scope of this pull request. It was just something I'd noticed, and thought it meant text could be drawn on images. Since that was the entirety of my evidence, it's very shaky evidence since it'd make much more sense to have border images anyway, and I'm lacking in knowledge in this field, I figure I'll just accept that text probably isn't drawn on top of images.

angstsmurf:
I looked into how Narcolepsy does it, and apparently it uses a non-standard WinGlulxe feature that gives the game window partial transparency by using an image as a mask. I don't think any other game or interpreter uses this, and it is not covered by the Glk spec.

Tynach:
@cspiegel, I can't get the code to link. Here's the command that's gotten me the farthest:

```sh
cc -O overlay.c -I/home/tynach/Software/Code/garglk/garglk -I/home/tynach/Software/Code/garglk/support/babel -L/home/tynach/Software/Code/garglk/build/linux.release/garglk -L/home/tynach/Software/Code/garglk/build/linux.release/babel -lgarglk -lgarglkmain -lfreetype -lpng -lm -ljpeg -lfontconfig -lSDL2 -lSDL2_mixer -lglib-2.0 -lgobject-2.0 -lcairo -lgtk-3 -lgdk-3
```

And here's the output:

```
/usr/bin/ld: /home/tynach/Software/Code/garglk/build/linux.release/garglk/libgarglk.a(cgfref.o): in function `glk_fileref_create_temp':
cgfref.c:(.text+0x217): warning: the use of `tempnam' is dangerous, better use `mkstemp'
/usr/bin/ld: /home/tynach/Software/Code/garglk/build/linux.release/garglk/libgarglk.a(babeldata.o): in function `gli_initialize_babel':
babeldata.c:(.text+0x61): undefined reference to `get_babel_ctx'
/usr/bin/ld: babeldata.c:(.text+0x6f): undefined reference to `babel_init_ctx'
/usr/bin/ld: babeldata.c:(.text+0x85): undefined reference to `babel_treaty_ctx'
/usr/bin/ld: babeldata.c:(.text+0x94): undefined reference to `release_babel_ctx'
/usr/bin/ld: babeldata.c:(.text+0xbf): undefined reference to `babel_treaty_ctx'
/usr/bin/ld: babeldata.c:(.text+0xd3): undefined reference to `release_babel_ctx'
/usr/bin/ld: babeldata.c:(.text+0xf4): undefined reference to `ifiction_get_tag'
/usr/bin/ld: babeldata.c:(.text+0x10f): undefined reference to `ifiction_get_tag'
```

Attempting to put `-lbabel`, `-lbabel_static`, and other variants leads to:

```
/usr/bin/ld: cannot find -lbabel
```

Where `-lbabel` changes depending on what I put after `-l`.

Tynach:
@cspiegel I've been chasing a red herring, caused by my misinterpretation of the code.

I still can't quite work out when one is called vs. the other, but it seems that images like PNG files are composited by either `gli_draw_picture()` in `draw.c`, or `drawpicture()` in `wingfx.c`. Neither one calls the gamma correction code, so the gamma correction code indeed only affects text rendering.. At least, the first time it's drawn.

Whenever `win_graphics_redraw()` (also in `wingfx.c`) is called, it ends up calling `gli_draw_pixel()` directly - which is where the gamma correction code is written. It's at that point that images get some color quality loss at dark levels (not noticeable on my system, but does throw off color pickers and the like when a color channel should have a value of 1 (on a 0-255 scale).

cspiegel:
@Tynach Do you want to continue investigating the PNG issue for this pull request, or would you rather get this merged in now?

Tynach:
@cspiegel That's up to you. I've graphed the current errors (or at least, what they mathematically should be) using Qalculate and GNUPlot:

gamma-error

**Note**: the default gamma of 1.0 eliminates all traces of the problem and always has. This will only show up, and only be visible in dark areas, if the user sets a custom gamma value.

The worst behavior occurs if you use a gamma value of 2.2 or higher, but even at 2.2 the 'highest' inaccuracy is that a value of 20 shows up as 21. The largest inaccuracy is that the values 0 to 5 all turn into 0, with a value of 6 turning into 8.

But using a gamma of 1.8 drops those inaccuracies significantly, and using the Rec. 709 curve eliminates the issue (with the current resolution of the inverse gamma lookup table).

**Note**: the default gamma of 1.0 eliminates all traces of the problem and always has. This will only show up, and only be visible in dark areas, if the user sets a custom gamma value.

Besides all that, I noticed that jpegs in particular seem to be drawn darker than they should be anyway. Don't know why, but I noticed this about Gargoyle before I touched the code. Dunno if it affects PNGs, either. Either way, there might be other issues with drawing images that are beyond the scope of this pull request.

What *might* be in scope of this code, is it might be worth having me turn this into a separate code path that gets called when drawing text, so that `win_graphics_redraw()` doesn't get affected. If we do that, then I can get away with lowering the resolution of the inverse gamma table again and just ignoring the inaccuracies (they won't be noticeable as they'll only affect a few pixels at a time in ways that are too subtle to notice).

But I'll leave that up to you guys. Personally I think it's weird to have blending code in 3 different functions, though it *did* prevent the potential bug of pngs being alpha blended in linear light, and it might also improve performance or something (not really sure, but it's plausible).

I also think I should have looked more into what code calls `gli_draw_pixel()` before modifying it, and been more careful about that. The bugs (inaccurate colors) aren't super severe this time around, but I really should have been more careful.

I think that ultimately, you guys probably know more about what is and isn't an acceptable tradeoff than I do, and are more familiar with the code than I am.

cspiegel:
> What _might_ be in scope of this code, is it might be worth having me turn this into a separate code path that gets called when drawing text, so that `win_graphics_redraw()` doesn't get affected. If we do that, then I can get away with lowering the resolution of the inverse gamma table again and just ignoring the inaccuracies (they won't be noticeable as they'll only affect a few pixels at a time in ways that are too subtle to notice).
> 
> But I'll leave that up to you guys. Personally I think it's weird to have blending code in 3 different functions, though it _did_ prevent the potential bug of pngs being alpha blended in linear light, and it might also improve performance or something (not really sure, but it's plausible).

If you don't mind doing that, I say go for it: cleaning up code plus improving behavior is a win.  But it's more work, so it's entirely up to you.  I'm also happy merging right now and then you can, if/when you desire, create a new merge request with the updated changes.  There's no harm in doing several different merges around the same area of code.

> I think that ultimately, you guys probably know more about what is and isn't an acceptable tradeoff than I do, and are more familiar with the code than I am.

At this point you're at least more familiar than I am!  The original author (Tor Andersson) handed off maintenance years ago to Ben Cressey, who, after several years, stopped having time for it, and so now I'm more of a de facto steward of the code than anything else. If somebody has taken the time to understand/modify a part of the code, I'll defer knowledge to them, such as I'm doing here.  At this point, you're the resident expert on gamma correction so however you'd like to proceed is up to you.

Tynach:
@cspiegel, well, I'm regretting having done some things in a hurry before. I hadn't first created a commit to fix only the blurry font bug, I didn't check to see if `gli_draw_pixel()` was called elsewhere, and I didn't even make a proper branch to work on the code.

I feel like what I *should* do is scrap this fork and pull request and start anew with a fresh one, making more incremental changes that are easier to roll back to and compare, especially as that would be nicer with things like the 'blame' tool. Thing is, I don't know how to do that with Github. Can I even delete the repository from my account and re-fork cleanly? Or does the fact that this pull request exists prevent that? Or will it delete this pull request and prevent us from referring back to this discussion?

cspiegel:
> I feel like what I _should_ do is scrap this fork and pull request and start anew with a fresh one, making more incremental changes that are easier to roll back to and compare, especially as that would be nicer with things like the 'blame' tool. Thing is, I don't know how to do that with Github. Can I even delete the repository from my account and re-fork cleanly? Or does the fact that this pull request exists prevent that? Or will it delete this pull request and prevent us from referring back to this discussion?

You can delete and refork, no problems.  That'll make this pull request invalid, but shouldn't delete it: it'll be here for posterity.  I can close it and you can open a fresh one based on your new, updated code, and you can refer to this pull request in the new one, so there's a link (just by doing something like "See pull request #419 for more information").

One suggestion: when you're working in your own forked repository, I'd still recommend using a separate branch so that your master is clean.  You can request merges from any branch to this repository with no problems.  You can also use a bit of Git magic to keep your current setup, switching to branching along the way. That way you don't need to delete and refork. The steps, with comments:

```
git checkout -b fix-gamma-old master # Create a new branch called "fix-gamma-old" based off of master
git remote add upstream https://github.com/garglk/garglk.git # Add a new remote called "upstream" which points to the garglk project
git fetch upstream # Get branch information from the new remote
git checkout master # Switch to your master branch
git reset --hard upstream/master # Reset your master branch to be in line with garglk master
```
At this point you now have a new branch called `fix-gamma-old` which is identical to your old branch, so you haven't lost any code.  Your `master` is now aligned with Gargoyle's master. At this point you can create a new branch off of master and re-apply any fixes you want.  If you'd like you can create one branch (and one pull request) for the Freetype fixes, and another branch/request for the gamma updates.

For each branch you can push to your own repository with:

```
git push -u origin branch-name
```
That will set things up so git pull/push will automatically use the remote branch with the same name.

If you lose your code, I have a copy of your branch locally here, so it's backed up in at least one place.

Tynach:
I started typing this about 15 minutes after your reply, but then life got in the way for a while. Below is mostly what I'd already typed up, with minimal editing.
___

> That'll make this pull request invalid, but shouldn't delete it: it'll be here for posterity.

Good to know. I think I'll do that.

> One suggestion: when you're working in your own forked repository, I'd still recommend using a separate branch so that your master is clean.

Yeah, that's what I regret not doing. Honestly, I mildly regretted it just while testing my code, as it was more difficult than necessary to checkout the last 'official' commit.

> The steps, with comments

I was already looking up how to do some of that stuff before checking for replies, and now I don't have to finish researching :) Thanks!

I didn't know for sure (but suspected; one of the things I hadn't yet looked up, but was planning to) if I could have different pull requests associated with separate branches of the same repo. Good to know for sure that's possible, and the `git push -u` bit was actually something I hadn't even thought of. I kinda thought pushing branches would be automatic, but I suppose it'd make sense for it not to be.

cspiegel:
> I was worried that gammacopy_lcd() was handling 3 color channels, and gammacopy() was handling only a single color channel. It's good to know I can just completely remove the _lcd() variant!

Be aware that I was doing things blindly, so I can't say one way or the other what the ultimate effects are of removing the lcd version. 

> Do you know if it's feasible to remove the copy entirely, and just use the original FreeType output? The new way of doing gamma correction will no longer be performing it in this 'copy' function, so the copy seems kinda pointless.

I don't, sorry; I hadn't ever looked at this code before you started working on it.  What I've noticeed is that the `gammacopy` function takes into account the gamma from `garglkrc`.  So, without it, no gamma correction is done (from what I can see a direct copy is equivalent to a gamma of 1.0).  But you probably are aware of that.  I mean, somebody has to do the gamma correction, and I'm not familiar with how that's "normally" done.

I wish I were more helpful here, but it's just not something I can speak on with any confidence.

> Somewhat unrelatedly, where should discussion about the next pull request go before the pull request is actually submitted? I have some questions about config options I'd like to discuss before making any commits in my second try, and it's not exactly relevant to the pull request fixing font filtering, nor this old one.

You can open a new issue for it.  Issues are really ad hoc here; there is an "enhancement" label that you can add since it's not a bug, but ultimately it doesn't really matter.  It's probably the best way to track things (and your pull request can reference an issue so there's a trail).

> I also posted this comment to the filtering pull request, but I'm uncertain how visible that'd be since the pull request was closed.

I definitely didn't see it there; tagging me here worked well since it resulted in an email sent.  Feel free to @ me in any comments, especially if it looks like I'm not responding.

Tynach:
> Be aware that I was doing things blindly, so I can't say one way or the other what the ultimate effects are of removing the lcd version.

It's fine this time. I've since discovered that `gammacopy()` - both the lcd and non-lcd versions - are just re-implementations of `memcpy()`, but with the values from the gamma lookup table being what gets written instead of the raw data. The only reason why the lcd version had to be different was because Gargoyle was performing its own filtering.

It's possible the code was written before FreeType offered similar five-tap filtering natively, but I'm not sure. I do know that FreeType couldn't enable the filtering by default for many years due to patents, but those have all expired now. At any rate, the API to choose different filters was added in version 2.3.0, back in January 2007.

The earliest commit for the filtering code in this repository is the initial import into *svn*, in 2008, by Ben Cressey. I don't know how old the code itself is, though; are there any ancient repositories of the code from before it was in either svn or git?

> I don't, sorry; I hadn't ever looked at this code before you started working on it. What I've noticeed is that the `gammacopy` function takes into account the gamma from `garglkrc`. So, without it, no gamma correction is done (from what I can see a direct copy is equivalent to a gamma of 1.0). But you probably are aware of that. I mean, somebody has to do the gamma correction, and I'm not familiar with how that's "normally" done.
> 
> I wish I were more helpful here, but it's just not something I can speak on with any confidence.

Hey, it's still helpful. I hadn't even thought to try before; I just wanted to make the minimum required changes. But now that I know that removing the gamma code just leaves me with a copy of `memcpy()`, I'll just be removing those functions entirely and using `memcpy()` directly (I've already tested to make sure it works).

I still feel like a copy operation shouldn't be necessary, but my attempts to just use the memory that FreeType already allocated result in.. Garbled nonsense. I mean, it compiles and runs, and images render fine, but text is absolutely garbled, though 'S'es tend to be barely legible for some unknown reason.

I can only assume that the secret snake government that mind controls the world's leaders has something to do with it.

I will say that I don't see where the memory gets freed, but I also don't think I see any obvious memory leaks. The calls to `free()` might be buried elsewhere in the code.

> You can open a new issue for it. Issues are really ad hoc here; there is an "enhancement" label that you can add since it's not a bug, but ultimately it doesn't really matter. It's probably the best way to track things (and your pull request can reference an issue so there's a trail).

I blame my lack of sleep the previous night for completely forgetting that 'enhancement' and other such issue tags exist. Will do!

> I definitely didn't see it there; tagging me here worked well since it resulted in an email sent. Feel free to @ me in any comments, especially if it looks like I'm not responding.

Fair enough :) You've been fairly responsive so far, especially given that you're probably busier in general than I am these days.

apodtele:
If you prefer your own filtering instead of current FreeType, you have to explicitly use FT_LCD_FILTER_NONE with FT_CONFIG_OPTION_SUBPIXEL_RENDERING. Note that *without* FT_CONFIG_OPTION_SUBPIXEL_RENDERING, filtering is not necessary on either side because the Harmony rendering (based on shifts) is already equivalent to the light filter. 

I would recommend relying of FreeType to do filtering or Harmony rendering.

Tynach:
@apodtele, all of what you've said was already addressed quite some time ago in my more recent set of more incremental patches in #433 (which is all about ripping out Gargoyle's own filtering and relying on FreeType's). Also, at no point was anyone saying that different gamma values should be used per channel.

I disagree that it even makes sense on a theoretical level; per-channel gamma correction is the operating system's job, via the use of VCGT (Video Card Gamma Table) values supplied by a display profile (or other file; I know ArgyllCMS has a utility for applying video card gamma tables via a space or comma separated value file, or something like that.. Haven't used that since I got an actual colorimeter and started making proper profiles for my displays).

Tynach:
Closing this, it's not relevant anymore.

PR #422, opened 2021-05-15
Adjust the SDK-checking glob in gargoyle_osx.sh

erkyrath:
Instead of an open-ended check for 10.14+, it now checks for SDK 10.7 through 10.13. Those get a minimum target of 10.7. Everything else gets a minimum target of 10.9.

(I could be finicky and fail with an error on 10.6 or earlier, but I don't think it's worth the effort.)


cspiegel:
Thank you!

PR #429, opened 2021-05-18
Fix fonts being subpixel filtered twice

Tynach:
Gargoyle does its own subpixel filtering of rendered fonts. This used to work fine with versions of FreeType prior to 2.10.3, but 2.10.3 and up use a default set of filter weights.

Previously, one would need to explicitly set the filter weights for any filtering to occur, leaving Gargoyle free to perform the filtering itself.

### DIFF
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -212,10 +212,12 @@ static void loadglyph(font_t *f, glui32 cid)
         if (f->make_oblique)
             FT_Outline_Transform(&f->face->glyph->outline, &ftmat);
 
-        if (gli_conf_lcd)
+        if (gli_conf_lcd) {
+            FT_Library_SetLcdFilter(ftlib, FT_LCD_FILTER_NONE);
             err = FT_Render_Glyph(f->face->glyph, FT_RENDER_MODE_LCD);
-        else
+        } else
             err = FT_Render_Glyph(f->face->glyph, FT_RENDER_MODE_LIGHT);
+
         if (err)
             winabort("FT_Render_Glyph");
 
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -31,6 +31,7 @@
 #include <ft2build.h>
 #include FT_FREETYPE_H
 #include FT_OUTLINE_H
+#include FT_LCD_FILTER_H
 
 #include <math.h> /* for pow() */
 #include "uthash.h" /* for kerning cache */


cspiegel:
I suppose the one question I have is, is it better for Gargoyle to do the filtering itself, or to let Freetype do it?  Or does your followup code expect Freetype _not_ to be doing any filtering?

I guess I'm just wondering if Gargoyle's doing extra, unnecessary work here.

cspiegel:
And perhaps I'm looking at the wrong thing, but I can't see a difference between this and how things look in master.  I tried with a gamma both of 1.0 and 1.8, and the output appears identical (I'm using Freetype 2.10.4).

However, if I use `FT_LCD_FILTER_DEFAULT` and then avoid using `gammacopy_lcd()` in `draw.c` (i.e. always just use `gammacopy()`, I do see an improvement in rendering: as with the previous pull request, it looks sharper.  There's a clear difference.

But I will readily admit I'm shooting in the dark here, so I may be completely misunderstanding things.

Tynach:
> I suppose the one question I have is, is it better for Gargoyle to do the filtering itself, or to let Freetype do it? Or does your followup code expect Freetype _not_ to be doing any filtering?
> 
> I guess I'm just wondering if Gargoyle's doing extra, unnecessary work here.

It's certainly better to let FreeType do the filtering, but with the way Gargoyle does its own filtering, the result is different (currently) if I were to let FreeType do the filtering.. But only because of when the gamma correction takes place. Presently, gamma correction is applied to the output of FreeType, before filtering takes place.

With proper gamma-correct alpha blending, the filtering has to take place first, and then the filtered result is used to blend between two values after those two values have gamma correction applied. Gamma correction is never applied to FreeType's output, only to the two colors being blended together (the background and the foreground). Afterward, the inverse gamma is applied to bring the colors back to display-space (they were blended in linear space).

However, this is a good first step because it introduces the line of code where FreeType's LCD filter will be set, so there will be fewer changes when the code is changed to let FreeType perform the filtering.

> And perhaps I'm looking at the wrong thing, but I can't see a difference between this and how things look in master. I tried with a gamma both of 1.0 and 1.8, and the output appears identical (I'm using Freetype 2.10.4).
> 
> However, if I use `FT_LCD_FILTER_DEFAULT` and then avoid using `gammacopy_lcd()` in `draw.c` (i.e. always just use `gammacopy()`), I do see an improvement in rendering: as with the previous pull request, it looks sharper. There's a clear difference.
> 
> But I will readily admit I'm shooting in the dark here, so I may be completely misunderstanding things.

It's an extremely subtle difference. Here's two pictures for comparison. First one is showing the bug, second one is showing the result after the fix, and both use a gamma of 1.0:

narco-gam10-old-bugfilt

narco-gam10-old-garfilt

But again, this is mostly just to put a commit in that allows for more easily comparing the output of the old and new code, and it also fixes a minor bug. In case there are further problems with the gamma correction code (maybe it causes some serious performance regression on some setups that we only find out about later on, for all we know), at least this gives a nice place to revert to.

Tynach:
I will say that the bug is more noticeable with a font that renders sharper in general. Here are two similar screenshots, this time using 'DejaVu Serif Condensed' instead of 'Noto Serif', at size 13.6:

narco-gam10-old-bugfilt-dvsfc

narco-gam10-old-garfilt-dvsfc

cspiegel:
Ok, got it. Thanks for the detailed explanation!

Tynach:
> However, if I use `FT_LCD_FILTER_DEFAULT` and then avoid using `gammacopy_lcd()` in `draw.c` (i.e. always just use `gammacopy()`), I do see an improvement in rendering: as with the previous pull request, it looks sharper. There's a clear difference.

I was worried that `gammacopy_lcd()` was handling 3 color channels, and `gammacopy()` was handling only a single color channel. It's good to know I can just completely remove the `_lcd()` variant!

Do you know if it's feasible to remove the copy entirely, and just use the original FreeType output? The new way of doing gamma correction will no longer be performing it in this 'copy' function, so the copy seems kinda pointless.

Somewhat unrelatedly, where should discussion about the next pull request go before the pull request is actually submitted? I have some questions about config options I'd like to discuss before making any commits, and it's not exactly relevant to *this* pull request, nor the old one.

PR #433, opened 2021-05-20
Remove Gargoyle's filtering in favor of FreeType's

Tynach:
Also adds two new configuration options, `lcdfilter` and `lcdweights`. See the commit's message for details; Github wants to format the ordered sub-lists with roman numerals when in this case it's important for them to be decimals. Rather than fight with Github's Markdown rendering weirdness, I'm just going to refer to the plaintext form that's in the commit message already.

One additional note about this commit: it does make text look significantly worse when using a gamma other than 1.0, since the 'gamma correction' is applied *after* filtering, rather than before. On the plus side, it makes text sharper (even with Gargoyle's filter weights).. But it also shows significantly worse color fringes.

This problem will be fixed once proper gamma-correct linear alpha blending is implemented in my next pull request.

### DIFF
--- a/garglk/config.c
+++ b/garglk/config.c
@@ -411,17 +411,23 @@ static void readoneconfig(char *fname, char *argv0, char *gamefile)
             gli_set_lcdfilter(arg);
 
         if (!strcmp(cmd, "lcdweights")) {
+            int i;
             char *start = arg;
             char *end;
+            unsigned char weights[5];
 
-            for (int i = 0; i < 5; ++i) {
-                gli_conf_lcd_weights[i] = strtoul(start, &end, 10);
+            for (i = 0; i < 5; ++i) {
+                weights[i] = strtoul(start, &end, 10);
 
                 if (start == end) {
                     break;
                 }
 
                 start = end;
+            }
+
+            if (i == 5) {
+                memcpy(gli_conf_lcd_weights, weights, sizeof weights);
             }
         }
 
--- a/garglk/config.c
+++ b/garglk/config.c
@@ -137,7 +137,6 @@ int gli_caret_shape = 2;
 int gli_link_style = 1;
 
 int gli_conf_lcd = 1;
-int gli_conf_lcd_filter = -1;
 unsigned char gli_conf_lcd_weights[5] = {28, 56, 85, 56, 28};
 
 int gli_wmarginx = 15;
@@ -409,7 +408,7 @@ static void readoneconfig(char *fname, char *argv0, char *gamefile)
             gli_conf_lcd = atoi(arg);
 
         if (!strcmp(cmd, "lcdfilter"))
-            gli_conf_lcd_filter = atoi(arg);
+            gli_set_lcdfilter(arg);
 
         if (!strcmp(cmd, "lcdweights")) {
             char *start = arg;
--- a/garglk/config.c
+++ b/garglk/config.c
@@ -137,6 +137,8 @@ int gli_caret_shape = 2;
 int gli_link_style = 1;
 
 int gli_conf_lcd = 1;
+int gli_conf_lcd_filter = -1;
+unsigned char gli_conf_lcd_weights[5] = {28, 56, 85, 56, 28};
 
 int gli_wmarginx = 15;
 int gli_wmarginy = 15;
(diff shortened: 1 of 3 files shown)


cspiegel:
I've got a suggested pull request available at Tynach/garglk#1.

cspiegel:
It turns out there's something of a sticky issue with Freetype (that you're probably aware of).

There are two subpixel rendering engines in Freetype: Cleartype-style, and Harmony.  Your changes affect the Cleartype-style rendering, but not Harmony.  Harmony is controlled with the `FT_Library_SetLcdGeometry` function.

It's fine just having Cleartype-style configuration right now, but it should be documented as such.  Which leads me to the point that the default `garglk.ini` should be updated to include and document these new options.  I'm fine adding them though I'm sure you'd be able to document them better than I can.

At some point it'd be nice to be able to let the user know if a configuration option doesn't work, e.g. if Harmony is the engine, but `lcdfilter` is used.  But that's true of _all_ config options, so isn't in scope here.

Tynach:
> It turns out there's something of a sticky issue with Freetype (that you're probably aware of).
> 
> There are two subpixel rendering engines in Freetype: Cleartype-style, and Harmony. Your changes affect the Cleartype-style rendering, but not Harmony. Harmony is controlled with the `FT_Library_SetLcdGeometry` function.

Indeed, though the Harmony engine does not have a customizable filter. At least, not in the same way. The `FT_Library_SetLcdGeometry()` function sets the positions of each subpixel, rather than setting how much neighboring subpixels contribute to the current subpixel's coverage weighting.

This means you can produce subpixel antialiasing for some of the less common subpixel patterns, as long as there's always one subpixel of each color channel per pixel (which isn't true of several variants of PenTile, but is with some, so it'd be possible to produce output with subpixel antialiasing for some of the more obscure OLED displays but not all of them).

However, Harmony antialiasing would not provide any means of reproducing how Gargoyle used to render text, instead looking roughly the same as using the 'light' ClearType-style LCD filter. In fact, Harmony also doesn't let you have no filter at all (you could put all subpixels in the same place, but that just gives you grayscale), so even without these changes we have to assume FreeType is providing ClearType-style antialiasing.

> It's fine just having Cleartype-style configuration right now, but it should be documented as such. Which leads me to the point that the default `garglk.ini` should be updated to include and document these new options. I'm fine adding them though I'm sure you'd be able to document them better than I can.

Gargoyle already depends on ClearType-style antialiasing filters for its own filtering, as there is no equivalent to `FT_LCD_FILTER_NONE` when using Harmony.

> At some point it'd be nice to be able to let the user know if a configuration option doesn't work, e.g. if Harmony is the engine, but `lcdfilter` is used. But that's true of _all_ config options, so isn't in scope here.

I don't think Harmony is used in any major Linux distros, especially since the patents holding back ClearType-style filtering have expired. For other platforms, I presume we can control what is and isn't included in the FreeType library shipped with the program.

This does make me curious if Gargoyle, since it implemented its own filtering before, was in violation of these patents before now.. Or if the patents covered something to do with the pre-filtering stage, that would produce the unfiltered output that Gargoyle applied its filtering to?

cspiegel:
> I don't think Harmony is used in any major Linux distros, especially since the patents holding back ClearType-style filtering have expired. For other platforms, I presume we can control what is and isn't included in the FreeType library shipped with the program.

Yes, for Windows, we can include a DLL with Cleartype-style rendering turned on: I just updated DLLs for Windows, including Freetype.  I'll have to go back and check whether it's using Cleartype or not.

Harmony configuration is a "nice for the future" thing, and can probably be done on-demand: if somebody runs into issues with an obscure display and they have Harmony enabled, I can look at supporting it then.

For now there's no problems with only supporting configuration for Cleartype.

cspiegel:
> This does make me curious if Gargoyle, since it implemented its own filtering before, was in violation of these patents before now.. Or if the patents covered something to do with the pre-filtering stage, that would produce the unfiltered output that Gargoyle applied its filtering to?

A good question that I don't know the answer to at all.  Thankfully it's not an issue we have to worry about anymore!

Tynach:
@cspiegel I did some research and found that Arch Linux used to use Harmony subpixel antialiasing by default, but in 2020 [switched to using ClearType-style rendering by default][arch]. There are a lot of mix-ups among the people posting in that bug report (confusion about subpixel filtering vs. hinting, among other things), but at the bottom it looks like it ended up being enabled by default after all.

[arch]: https://bugs.archlinux.org/task/60658

PR #435, opened 2021-05-27
Accurate gamma correction when rendering text

Tynach:
I've tried going with the 'alternate versions of functions' route this time, so that png and jpeg files are never rendered with inaccurate colors (at least, not due to my code here). I've also added some modifications to the configuration files, in particular adding documentation to the options added from #433.

I wasn't sure if `gamma` should be moved into the new section I made or not, and I figured before making that drastic of a change I'd get feedback on the modifications I'd already made.

**Edit**: Also, I should mention I've had a few ideas about how to allow someone to type in `srgb` or `rec709` instead of a number for the gamma value. Felt more important to get the code working without further configuration additions, though.

### DIFF
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -91,7 +91,7 @@ struct font_s
  */
 
 static unsigned short gammamap[256];
-static unsigned short gammainv[256 << GAMMA_SHIFT];
+static unsigned char gammainv[256 << GAMMA_SHIFT];
 
 static const int gammamax = (256 << GAMMA_SHIFT) - 1;
 
--- a/garglk/draw.c
+++ b/garglk/draw.c
@@ -37,7 +37,10 @@
 #include <math.h> /* for pow() */
 #include "uthash.h" /* for kerning cache */
 
+#define GAMMA_SHIFT 3
+
 #define mul255(a,b) (((a) * ((b) + 1)) >> 8)
+#define mulhigh(a,b) (((a) * ((b) + 1)) >> (8 + GAMMA_SHIFT))
 #define grayscale(r,g,b) ((30 * (r) + 59 * (g) + 11 * (b)) / 100)
 
 #ifdef _WIN32
@@ -87,7 +90,10 @@ struct font_s
  * Globals
  */
 
-static unsigned char gammamap[256];
+static unsigned short gammamap[256];
+static unsigned short gammainv[256 << GAMMA_SHIFT];
+
+static const int gammamax = (256 << GAMMA_SHIFT) - 1;
 
 static font_t gfont_table[8];
 
@@ -148,12 +154,6 @@ static int touni(int enc)
         case ENC_MDASH: return UNI_MDASH;
     }
     return enc;
-}
-
-static void gammacopy(unsigned char *dst, unsigned char *src, int n)
-{
-    while (n--)
-        *dst++ = gammamap[*src++];
 }
 
 static int findhighglyph(glui32 cid, fentry_t *entries, int length)
@@ -226,7 +226,7 @@ static void loadglyph(font_t *f, glui32 cid)
         glyphs[x].h = f->face->glyph->bitmap.rows;
         glyphs[x].pitch = f->face->glyph->bitmap.pitch;
         glyphs[x].data = malloc(datasize);
-        gammacopy(glyphs[x].data, f->face->glyph->bitmap.buffer, datasize);
+        memcpy(glyphs[x].data, f->face->glyph->bitmap.buffer, datasize);
     }
 
     if (cid < 256)
@@ -384,7 +384,10 @@ void gli_initialize_fonts(void)
     int i;
 
     for (i = 0; i < 256; i++)
-        gammamap[i] = pow(i / 255.0, gli_conf_gamma) * 255.0;
+        gammamap[i] = pow(i / 255.0, gli_conf_gamma) * gammamax + 0.5;
+
+    for (i = 0; i <= gammamax; i++)
+        gammainv[i] = pow(i / (float)gammamax, 1.0 / gli_conf_gamma) * 255.0 + 0.5;
 
     err = FT_Init_FreeType(&ftlib);
     if (err)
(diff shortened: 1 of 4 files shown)


Tynach:
@cspiegel I've written a couple of variations of how to incorporate the sRGB and Rec. 709 transfer functions. In my opinion, the second one is better - but the first one has simpler option parsing and doesn't introduce additional functions into the codebase. The second one introduces an additional struct type, 2 new functions, and slightly more complicated configuration parsing.

## Enum Selection

The first uses a combination of an `enum` with the different options, and a function similar to what you wrote for parsing the LCD filter presets that assigns an instance of the enum, and that's used to pick whether to use a straight power function or a predefined curve like sRGB. It also uses this mess of `if` statements and manually typed out formulas at the top of `gli_initialize_fonts()`:

```c
if (gamma_type == POWER) {
    for (i = 0; i < 256; i++) {
        gammamap[i] = pow(i/255.0, custom_gamma)*gammamax + 0.5;
    }

    for (i = 0; i <= gammamax; i++) {
        gammainv[i] = pow(i/(float)gammamax, 1.0/custom_gamma)*255.0 + 0.5;
    }
} else if (gamma_type == SRGB) {
    for (i = 0; i < 256; i++) {
        if (i/255.0 <= 0.04045) {
            gammamap[i] = i/255.0/12.92*gammamax + 0.5;
        } else {
            gammamap[i] = pow((i/255.0 + 0.055)/1.055, 2.4)*gammamax + 0.5;
        }
    }

    for (i = 0; i <= gammamax; i++) {
        if (i/(float)gammamax <= 0.0031308) {
            gammainv[i] = i/(float)gammamax*12.92*255.0 + 0.5;
        } else {
            gammainv[i] = (1.055*pow(i/(float)gammamax, 1.0/2.4) - 0.055)*255 + 0.5;
        }
    }
} else if (gamma_type == REC709) {
    for (i = 0; i < 256; i++) {
        if (i/255.0 <= 0.081) {
            gammamap[i] = i/255.0/4.5*gammamax + 0.5;
        } else {
            gammamap[i] = pow((i/255.0 + 0.099)/1.099, 1.0/0.45)*gammamax + 0.5;
        }
    }

    for (i = 0; i <= gammamax; i++) {
        if (i/(float)gammamax < 0.018) {
            gammainv[i] = i/(float)gammamax*4.5*255.0 + 0.5;
        } else {
            gammainv[i] = (1.099*pow(i/(float)gammamax, 0.45) - 0.099)*255 + 0.5;
        }
    }
}
```

This function is used to process the `gamma` value in this first variant, and I do like it's simplicity:

```c
void gli_set_gamma(const char *gamma_input)
{
    if (strcmp(gamma_input, "srgb") == 0) {
        gamma_type = SRGB;
    } else if (strcmp(gamma_input, "rec709") == 0) {
        gamma_type = REC709;
    } else {
        gamma_type = POWER;
        custom_gamma = atof(gamma_input);

        // A gamma of 0 does not make sense, nor does a negative gamma
        if (custom_gamma <= 0.0) {
            custom_gamma = 1.0;
        }
    }
}
```

And this is the (very simple) enum it uses:

```c
enum gamma_curve_e
{
    POWER = 0,
    SRGB = 1,
    REC709 = 2
};
```

Though, it needs both an instance of the enum *and* a float (`custom_gamma`; not terribly happy with the name, but it works for now) as global variables to select which is used, rather than a single contained variable, like the second option.

## Curve Struct

The second variation is one I like much better. It mostly just reimplements how I do it in my shaders (having a couple of separate functions that do the conversion, and having those functions use values stored in a `struct` that holds the curve's parameters). That mess of `if` statements is boiled down to something similar to what was there before, seen here:

```c
for (i = 0; i < 256; i++) {
    gammamap[i] = tolinear(i/255.0, gamma_curve)*gammamax + 0.5;
}

for (i = 0; i <= gammamax; i++) {
    gammainv[i] = togamma(i/(float)gammamax, gamma_curve)*255 + 0.5;
}
```

Though it does have some additional functions, which for now I include right above `gli_initialize_fonts()`. I'm not sure if the file should be organized some other way instead though, given that there are a few 'header' comments that look like they're meant to break the file up into sections. These are the functions themselves:

```c
static float tolinear(float color, transfer_curve_t trc)
{
    if (color <= trc.cutoff_linear) {
        return color/trc.slope;
    } else {
        return pow((color + trc.offset)/(1 + trc.offset), trc.power);
    }
}

static float togamma(float color, transfer_curve_t trc)
{
    if (color <= trc.cutoff_gamma) {
        return color*trc.slope;
    } else {
        return (1 + trc.offset)*pow(color, 1/trc.power) - trc.offset;
    }
}
```

At the top of the file I declare the `transfer_curve_t` typedef, and a corresponding `transfer_curve_s` structure, keeping with the style of the rest of the file. The struct is pretty simple:

```c
struct transfer_curve_s
{
    float power;
    float offset;
    float slope;
    float cutoff_linear;
    float cutoff_gamma;
};
```

However, the function for processing the gamma input string is slightly more complicated:

```
void gli_set_gamma(const char *gamma_input)
{
    float gamma;

    if (strcmp(gamma_input, "srgb") == 0) {
        gamma_curve = (transfer_curve_t){2.4, 0.055, 12.92, 0.04045, 0.0031308};
    } else if (strcmp(gamma_input, "rec709") == 0) {
        gamma_curve = (transfer_curve_t){1.0/0.45, 0.099, 4.5, 0.0812, 0.018};
    } else {
        gamma = atof(gamma_input);

        // A gamma of 0 does not make sense, nor does a negative gamma
        if (gamma > 0.0) {
            gamma_curve = (transfer_curve_t){gamma, 0, 1, 0, 0};
        } else {
            gamma_curve = (transfer_curve_t){1, 0, 1, 0, 0};
        }
    }
}
```

Though in my opinion it's also more extensible, and avoids having more than one global variable containing gamma curve parameters (even a pure power function can be expressed, as shown within the bottom `if` statement).

cspiegel:
> The second variation is one I like much better.

That says enough for me; you're the implementor, so if it's better for you, that's fine.

> and slightly more complicated configuration parsing.

It sounds like you're happy with the tradeoff of more complicated parsing vs the rest of the code, so no problems here.

> Though it does have some additional functions, which for now I include right above gli_initialize_fonts(). I'm not sure if the file should be organized some other way instead though, given that there are a few 'header' comments that look like they're meant to break the file up into sections.

There's not any sort of set style guide for function layout.  There should be, but it's just been ad hoc so far.  Put them where they feel reasonable, and that's fine.  At some point (as with so many things) I want to do some organizational work on the codebase, but it's a lower priority.

cspiegel:
Looks good on Windows, too.  Thanks!

Tynach:
@cspiegel, I should have probably worded 'more complicated parsing' differently. Being more awake today, I'm realizing that's not actually what I meant; it's more that the act of setting the variable value is using a struct that has little to no documentation for what the parameters mean, so it looks like a bunch of voodoo wizardry (especially for sRGB and Rec. 709). Basically, `{2.4, 0.055, 12.92, 0.04045, 0.0031308}` probably looks like nonsense if you're trying to figure out what the code actually *does*.

> There's not any sort of set style guide for function layout. There should be, but it's just been ad hoc so far. Put them where they feel reasonable, and that's fine. At some point (as with so many things) I want to do some organizational work on the codebase, but it's a lower priority.

Huh, it seemed pretty consistent to me. Public things (functions and variables available to other source files elsewhere in the project) use underscores between words, private things (`static` functions and globals; things specific to the file and not in any headers) run words together.

Though I sorta botched that with my `_gamma` pixel drawing function variations. I made those `static` and didn't add them to the headers, but I kept the underscores.

On that note, I feel kinda strange about having separate `_lcd()` variants, when those could be combined into a single function (and when using grayscale antialiasing, simply create an array of 3 of the same value and pass that in as the alpha instead of having a separate non-lcd function). But, I figured I'd first just stick with the way it's currently organized and not accidentally break anything this time around; refactors can be done separately, after all.

Issue #434, opened 2021-05-25
Visual artifacts and glitches


speachy:
I hope this is a case of PEBKAC, but ever since the port to GTK3 landed (commit 9a283b1734c1fd279da19b27c948ba458560b90d) garglk has been unusable for me:

 * No window decorations (titlebar, window borders, resize handles, can't move window, etc)
 * Scrolling/repainting artifacts (outer edge of window is not repainted properly when scrolling, leading to artifacts)
 * Typed text does not appear until enter is pressed

This seems to happen regardless of which engine is used (zcode, tads, adrift)

I'm compiling from source, running on Fedora 34 (Gnome 40, GTK 3.24.29, Wayland)

Thanks!

speachy:
![Screenshot from 2021-05-25 22-01-18](https://user-images.githubusercontent.com/5199155/119591591-e3426e80-bda4-11eb-9536-acd7f0c704e3.png)

Here's an example, running I-0.z5.  Launch, press any key to continue, and it's clear the scrolling/pagination is shot.  Plus the lack of window decoration.  garlglk freshly compiled from today's sources.

Tynach:
I don't have that issue, but I'm using KDE Neon. I do know that Gnome, when running GTK apps at least, sometimes assumes that the GTK apps themselves are going to draw their own window decorations. Perhaps it's detecting the app uses GTK3 and makes that assumption?

Does Qalculate also fail to draw window decorations? It's a GTK3 app that doesn't typically draw its own window decorations, at least on KDE and such; maybe it's smarter about that and draws it under Gnome, though.

speachy:
I think the problem is that newer versions of GTK3 (3.12, ~2015) default to using client-side decorations.  When running under a native GTK-based desktop (ie gnome) the desktop expects the application to handle all of that stuff.  When running standalone, the decorations end up provided by the system.

Tynach:
@speachy, does it also happen in KDE under wayland?

speachy:
I don't have any systems with KDE installed; the best I could do is spin up a VM and see what happens, but I'm not likely to have the bandwidth for that for a few days.

Tynach:
[...]

PR #442, opened 2021-06-10
Correct visual glitches and functional regressions when running under Gnome+Wayland (with CSD)

speachy:
This patch series resolves all of the visual glitches and functional regressions described in #434 when running under Gnome+Wayland, with no new regressuins under Gnome+X11

Untested under KDE and other non-Gnome environments.

### DIFF
--- a/garglk/sysgtk.c
+++ b/garglk/sysgtk.c
@@ -556,30 +556,29 @@ void winopen(void)
 
     frame = gtk_window_new(GTK_WINDOW_TOPLEVEL);
     gtk_widget_set_can_focus(frame, TRUE);
-    gtk_widget_set_events(frame, GDK_BUTTON_PRESS_MASK
-                               | GDK_BUTTON_RELEASE_MASK
-                               | GDK_POINTER_MOTION_MASK
-                               | GDK_POINTER_MOTION_HINT_MASK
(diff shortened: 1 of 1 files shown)


garglk/sysgtk.c
```diff
@@ -608,12 +611,6 @@ void winopen(void)
 
     wintitle();
 
-    gtk_window_set_geometry_hints(GTK_WINDOW(frame),
```
cspiegel, line 611: The only regression I'm seeing in KDE is that if the window is too large (i.e. cols and/or rows cause it to overflow the screen), the window manager doesn't/can't resize it to fit.  I don't think one way is _better_ than the other, but I'd prefer to keep the previous behavior for continuity, if possible.

I did some small tests and it's easy to get it back working in KDE.  Include `gdk/gdkwayland.h` and reinstate this deleted code, protected by `if gtk_window_set_geometry_hints(GTK_WINDOW(frame)`; i.e. run this only under non-Wayland.

Then, above, call `gtk_widget_set_size_request(canvas, defw, defh)` only if it _is_ Wayland.  The rest of the changes are fine in KDE.

If that works under Wayland with Gtk then we should be good to merge.

cspiegel:
Closing due to the conversion to Qt

PR #443, opened 2021-06-11
Switch the Unix UI to Qt

cspiegel:


cspiegel:
As of right now you have to manually specify the `C++` Jam variable: Jam apparently sets `C++` to `cc`, which generally works fine for compiling (as the compiler driver chooses which backend to use based on the file's extension), but it breaks for linking, since the C compiler doesn't know to link in, at least, the standard C++ library.  So the linker is set to `$(C++)`, but since that's `cc` by default, it won't work.

In addition, the build system assumes Qt's `moc` is called `moc-qt5`.  This can be changed by setting `MOC`.  Example:

```
jam -sCC=clang -sC++=clang++ -sMOC=moc
```
There's probably a better way to do this with Jam, but CMake support is coming, which will make life much easier, so Jam will contunue to run with duct tape and baling wire for the time being.

cspiegel:
@cibersheep It's been almost a year since you asked about this, but I've finally gotten it into decent shape.

cibersheep:
> @cibersheep It's been almost a year since you asked about this, but I've finally gotten it into decent shape.

I'm overexcited ^__^ willing to test as soon as I can.
Thanks for pinning me

cspiegel:
> It seems that both -sC++ and -sCXX need to be set to clang++ to get it to work.

More gifts from Jam (I believe it's just that CC and C++ both are needed, although for the life of me I don't know why).  These frustrations will be a thing of the past as soon as CMake support is merged in.  Thanks for persevering through the transition here.  I'll hope to have CMake in within the next couple of days.  I've decided it's worth having two build systems for the time being, since the benefits of CMake are so great.

> However, not all games seem to work properly in the resulting binaries of git. I'm not 100% sure why, but I suspect the game in question uses timed events. Normally it would show some text, then moments later show more text.. Or sometimes it would show an animation (running fairly smoothly too). Both sorts of things seem to just cause it to 'hang', sort of.. I can select text and whatnot, but the animation frame doesn't update, and new text does not appear. Resizing the window, ironically, causes the game to finally move on.

I tested timing with TerpEtude while implementing the Qt interface, but that's only making use of the Z-machine's limited use of timers (and then, only Bocfel's implementation of it).  Turns out I didn't properly implement them.  I've just pushed a fix for that.  See if it addresses the issue you're seeing.

> Also, glulxe simply segfaults no matter what.

I can't reproduce this.  Gluxe is working fine here (and glulxercise passes with flying colors).  Any chance you could provide a backtrace from gdb?

Tynach:
> More gifts from Jam (I believe it's just that CC and C++ both are needed, although for the life of me I don't know why).

Not here. If I only define `-sCC` and `-sC++`, some things fail to compile. It seems that some things still try to compile with GCC (the default for my system), while others compile with Clang. Here's what it prints for one of them (there are others, with longer but similar output):

```
Link build/linux.release/geas/geas 
/usr/bin/ld: build/linux.release/geas/geasfile.o: relocation R_X86_64_32 against symbol `_ZSt4cerr@@GLIBCXX_3.4' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: build/linux.release/geas/geas-runner.o: relocation R_X86_64_32 against symbol `_ZSt4cerr@@GLIBCXX_3.4' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: build/linux.release/geas/geas-state.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: build/linux.release/geas/geas-util.o: relocation R_X86_64_32 against symbol `_ZSt4cerr@@GLIBCXX_3.4' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: build/linux.release/geas/geasglk.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: build/linux.release/geas/geasglkterm.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: build/linux.release/geas/istring.o: relocation R_X86_64_32 against symbol `_ZSt4cerr@@GLIBCXX_3.4' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: build/linux.release/geas/readfile.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE

g++  -o build/linux.release/geas/geas  build/linux.release/geas/geasfile.o build/linux.release/geas/geas-runner.o build/linux.release/geas/geas-state.o build/linux.release/geas/geas-util.o build/linux.release/geas/geasglk.o build/linux.release/geas/geasglkterm.o build/linux.release/geas/istring.o build/linux.release/geas/readfile.o build/linux.release/garglk/libgarglkmain.a build/linux.release/garglk/libgarglk.so -lm `pkg-config freetype2 fontconfig Qt5Core Qt5Widgets libpng zlib sdl2 SDL2_mixer --libs`

...failed Link build/linux.release/geas/geas ...
...skipped <installed>geas for lack of geas...
```

> These frustrations will be a thing of the past as soon as CMake support is merged in. Thanks for persevering through the transition here. I'll hope to have CMake in within the next couple of days. I've decided it's worth having two build systems for the time being, since the benefits of CMake are so great.

Lets hope so!

> I tested timing with TerpEtude while implementing the Qt interface, but that's only making use of the Z-machine's limited use of timers (and then, only Bocfel's implementation of it). Turns out I didn't properly implement them. I've just pushed a fix for that. See if it addresses the issue you're seeing.

Seems to work, thanks!

> I can't reproduce this. Gluxe is working fine here (and glulxercise passes with flying colors). Any chance you could provide a backtrace from gdb?

I can't figure out how to make a debug build with Jam. I can only seem to find documentation for changing the debugging level of Jam itself, to show things about Jam's own internal state at each step.

Tynach:
@cspiegel, forgot to mention you in that post. Which is just as well, given that the new information I edited in at the end kinda deserves its own post.

Debug builds of `glulxe` work fine, but release builds do not.

cspiegel:
@Tynach I just fixed a (possibly crashing) bug that may be the source of your `glulxe` error.  Please give it a try and let me know.

As for backtraces, if there's still a crash with current master, I'll sync the cmake branch up with master and if you're willing, you can give a shot with that, since it will be much easier to do a debug build.

Tynach:
@cspiegel, that seems to do the job, glulxe is no longer segfaulting on startup.

Just to clarify though, I had already manage to make a debug build of glulxe before, and it worked fine. It was only release builds that were segfaulting on startup.

Is it possible that there are various environment variables in your setup that are different than I'd have? Maybe an environment variable somewhere is being used when calculating the parameters of a releasea build that are not used when calculating the parameters of a debug build (instead other values being manually set), and it turns out that the values used in your particular setup are necessary to make it work?

For example, I think one could probably set `CC` and `CXX` using environment variables instead of `jam` parameters; perhaps you have `CXX` set as an environment variable, and that's why you don't need to set `-sCXX=clang++` like I do.

PR #444, opened 2021-06-15
Add CMake support

cspiegel:


curiousdannii:
Can you build garglk as a static library with these Cmake definitions?

cspiegel:
> Can you build garglk as a static library with these Cmake definitions?

Not currently, but that's a great suggestion.  I'll add support for static library.

cspiegel:
OK, `-DBUILD_SHARED_LIBS=OFF` will now build a static `libgarglk`.

curiousdannii:
Fantastic! Thanks :)

curiousdannii:
While I'm here dropping wishes, how difficult would it be to add a setting to only build the garglk library? Or is that not even necessary, if we just build the garglk/ folder directly rather than the toplevel CMakeLists.txt?

cspiegel:
@curiousdannii I've just added two flags: `WITH_INTERPRETERS` and `WITH_LAUNCHER`.  If those are set to false, then only the library will be built.

Issue #455, opened 2021-06-21
Unix - allow running gargoyle from the build dir without installing


salty-horse:
To run gargoyle from the build directory (without running `make install`), I found that I have to edit `GARGLK_INTERPRETER_DIR` in `garglk/CMakeLists.txt`.

It is reasonable to change gargoyle to consult a `GARGLK_INTERPRETER_DIR` environment variable before using the hard-coded value?
Or, even to look in `terps/` in the current directory as a first attempt.

cspiegel:
A very reasonable request.  I've got a merge request at #458 to implement this (specifically, to look up the directory from the environment variable `GARGLK_INTERPRETER_DIR`).  Please give it a test to see if it suits your needs.

Note that this *replaces* the normal lookup, so if the variable is set but no interpreters are there, Gargoyle will not fall back to the system directory.  This is largely because the code to find and load interpreters isn't particularly suited to such a modification, at least until it's been cleaned up some.

salty-horse:
Thanks, it works! The new detailed error message when the interpreter can't be found is also useful.

I would suggest adding it to the `INSTALL` documentation

salty-horse:
Looks good

cspiegel:
Merged.

PR #458, opened 2021-06-22
Allow the interpreter directory to be overridden at runtime

cspiegel:
See #455.


cspiegel:
> This can be rewritten to avoid calling `getenv` twice

Busted!  I thought about that as I was writing the call out twice, but I just left it as-is.  You're right, of course, so I'll change it.

cspiegel:
@salty-horse OK, new addition to _not_ call it twice.  I'd also squashed an update to `INSTALL`, as you recommended, in preparation for merge.

salty-horse:
The change works fine :)

Issue #461, opened 2021-06-25
Add icons for mimetypes


jusw85:
Add icons to mimetypes for file manager integration.

e.g. in interactive-fiction.xml:
```
  <mime-type type="application/x-adrift">
    <comment>Adrift game data</comment>
    <magic priority="50">
      <match type="string" offset="0" value="\x3C\x42\x3F\xC9\x6A\x87\xC2\xCF"/>
    </magic>
    <glob pattern="*.taf"/>
    <icon name="application-x-gargoyle"/>
  </mime-type>
```

As per [FreeDesktop Icon Naming Specification](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-0.8.html#mimetypes), I would recommend renaming `gargoyle-docu2.png` to `application-x-gargoyle.png`

cspiegel:
I agree that adding an icon has utility; I'm a bit unsure how this ought to be done, though.

Ideally `interactive-fiction.xml` wouldn't be handled by Gargoyle at all.  It's more generic than that (and as a note, it overlaps with a MIME package provided by QTads, which naturally takes over the TADS file types).  I think the best thing would be to get this integrated into `shared-mime-info` although I don't know the process to do that; do you have any ideas?  The real problem, I think, is that this hasn't been standardized at all, so we don't even have real MIME types: I just made ones up so _something_ could be assigned in the `.desktop` file.

I'm fine updating this for now, but ultimately it'd be best to integrate it elsewhere.  I just haven't put any effort into doing so.

jusw85:
After further investigation, I realise this change isn't needed - the [default icon name](https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-0.21.html#idm46070616650832) can be defined by replacing '/' with '-'

```
Generally the icon used for a mimetype is created based on the mime-type by mapping "/" characters to "-", but users can override this by using the icon element to customize the icon for a particular mimetype.
```

I can install the icon as `application-x-adrift.png`, `application-x-blorb.png`, etc. in `/usr/share/icons/hicolor/32x32/mimetypes`. After updating the icon cache, it should be picked up by competent desktops / file managers (it does for mine).

[...]

PR #462, opened 2021-06-25
Add icons to mime specification

jusw85:
See #461. For integration with desktops and file managers


cspiegel:
The AppImage action also uses this icon, so it'll have to be updated with the new name.

PR #468, opened 2021-07-06
Use the STL instead of custom data structures

cspiegel:
This necessitates a switch to C++ for this file. At the present moment
it's still a mix of C- and C++-style code, with the latter being used so
that the STL's map can be used instead of custom data structures.

Rationale: Custom data structures are more error-prone than the STL, and
make the code harder to read. In addition, a bundled hash implementation
(that could fall out of date easily) is no longer needed. The less code
Gargoyle itself has, the better.


cspiegel:
@Tynach Does this clash with anything you're currently working on?  If there's anything in the near future you'd want merged into `draw.c` I'd rather get that in before this so I'm the one dealing with merge conflicts instead of you.

Tynach:
@cspiegel, I'll be honest, I sorta went on a spiral of looking at other mildly related topics and have a bunch of various diffs saved instead of branches, because I'm constantly reverting all changes and then re-doing relatively minor sets of changes in different combinations. It's very suboptimal, but I have no idea which things I'd ever even want to present to the rest of you.

For instance, I realized that adding sRGB and Rec709 support, while cool, wouldn't be of extreme benefit.. And it's one case where case sensitivity really can matter. When I added it locally, I kept accidentally spelling it as `sRGB` in my configuration file and it wouldn't work as a result, and that's kinda a problem. Sure there are ways of doing the check without case sensitivity, but since it'd be of little benefit anyway, I just figured I'd ask once I had other things sorted out.

Speaking of other things to sort out, boy did my exploration wander about with this. I did end up submitting that one change where I made some of the math more precise (#460), but that was originally part of a much larger set of changes that changed the blending algorithm to basically be `mul255(rgb, alpha) + mul255(p, invalf)` instead of `rgb + mul255(p - rgb, invalf)`. I had read somewhere that this was a more stable method. However, further investigation showed that the primary benefit was when dealing with floating point numbers, which isn't the case here.. And it *does* involve a higher number of operations being performed, so potentially performs worse.

Relatedly, I found that setting `GAMMA_BITS` to 16 with the current blending method resulted in garbage output, but worked just fine if I used the above blending method. However, I *later* found that I could get it to work just fine with the current/faster blending math by just adding some typecasts. But that's only necessary if `GAMMA_BITS` is set to 16, and I'm not sure if that's something you'd want to do or not.

Aand finally, I uh.. Started experimenting a *lot* with font hinting. The trouble there? There are a ton of different ways to do this, and doing it properly can occasionally mean letting go of subpixel positioning, unless you do some other various things like grabbing the advance width from `f->face->glyph->linearHoriAdvance` instead of `f->face->glyph->advance.x`, and of course then shifting over by 16 bits instead of 6 bits. Which FreeType has a handy function available for doing, `FT_MulFix()`. But if you try to use overly-sharpened grid fitting for the hinting (this is done by ORing the glyph's load flags with `FT_LOAD_TARGET_MONO`, but setting the render mode to `FT_RENDER_MODE_LCD` or `..._LIGHT`), this backfires tremendously since all but the first glyph will be very blurry. If we then backtrack and use `advance.x` again, there are *still* problems: glyphs that come after a glyph that was kerned will be very blurry, while glyphs before the kerned glyph will still be sharp.

This led me to discovering an odd problem that I'm not sure anyone here has noticed yet:

Many fonts are not properly kerned, and this is *not* Gargoyle's fault at all. FreeType does not support the GPOS table, which most OpenType fonts use to specify kerning, and an increasing number of TrueType fonts are using as well for kerning... Including the copy of Noto Serif that is currently included in FreeType. This means that a default copy of Gargoyle does not have kerning.

It turns out that the GPOS table can contain very advanced kerning instructions that allow kerning to depend on multiple glyphs surrounding the kerned glyph in question, so that contextual kerning can be done intelligently. Since FreeType only supports kerning of sets of kerned glyph pairs, GPOS is considered out of scope. Instead, Harfbuzz is recommended - and it's a library that can also support things like a font's internal custom ligatures and whatnot. Basically it does a lot of things Gargoyle currently does itself, and probably does a better job. We might want to use it.

Yesterday I was recovering from my second vaccination shot, so I apologize for not seeing your post until today. But overall I've just been kinda feeling overwhelmed (I know, entirely my own fault for taking on a bunch of ideas simultaneously). I'm not really sure what, if anything, should be changed.

Tynach:
Oh yeah, I also started messing with the initialization of `gli_image_rgb` to use 3 RGB subpixels instead of 4, using `QImage::Format_RGB888` so that the underlying pixel format matches what FreeType outputs. I got all that working locally, and the next step there is to rewrite the `draw_pixel` functions to just operate on one subpixel at a time, letting the `draw_bitmap` functions loop over all the subpixels instead of all the pixels.

However, while this would greatly simplify writing and maintaining these functions... I'm not sure I can really advocate for this. I'd kinda like to see that one feature of Windows Git/Glulxe that lets Narcolepsy define a transparent area of the window be implemented, but for that to happen we'd need that fourth subpixel so we can tell Qt which parts of the window to make transparent. I've seen semi-transparent sections of windows in Qt applications before, so I *know* it's possible.

I think that's all the different things I've been tinkering with.

cspiegel:
> @cspiegel, I'll be honest, I sorta went on a spiral of looking at other mildly related topics and have a bunch of various diffs saved instead of branches, because I'm constantly reverting all changes and then re-doing relatively minor sets of changes in different combinations. It's very suboptimal, but I have no idea which things I'd ever even want to present to the rest of you.
> 
> For instance, I realized that adding sRGB and Rec709 support, while cool, wouldn't be of extreme benefit.. And it's one case where case sensitivity really can matter. When I added it locally, I kept accidentally spelling it as `sRGB` in my configuration file and it wouldn't work as a result, and that's kinda a problem. Sure there are ways of doing the check without case sensitivity, but since it'd be of little benefit anyway, I just figured I'd ask once I had other things sorted out.

As somebody who has a bunch of branches hanging around with half-finished ideas, I know where you're coming from!
 
> Speaking of other things to sort out, boy did my exploration wander about with this. I did end up submitting that one change where I made some of the math more precise (#460), but that was originally part of a much larger set of changes that changed the blending algorithm to basically be `mul255(rgb, alpha) + mul255(p, invalf)` instead of `rgb + mul255(p - rgb, invalf)`. I had read somewhere that this was a more stable method. However, further investigation showed that the primary benefit was when dealing with floating point numbers, which isn't the case here.. And it _does_ involve a higher number of operations being performed, so potentially performs worse.
> 
> Relatedly, I found that setting `GAMMA_BITS` to 16 with the current blending method resulted in garbage output, but worked just fine if I used the above blending method. However, I _later_ found that I could get it to work just fine with the current/faster blending math by just adding some typecasts. But that's only necessary if `GAMMA_BITS` is set to 16, and I'm not sure if that's something you'd want to do or not.

It all depends on the utility.  If there are material improvements to be made, I'm all for it.  As for performance issues, the only real test is profiling.  There's no hard cutoff, but if (for example) there's a difference of a few percent in performance, it's probably worth it for better results.  I don't know how much effort has been given in the past to performance tuning, but there is likely a lot of stuff going on that could be optimized.  That may be something I look at in the near future, as there may be some low-hanging fruit.

> Aand finally, I uh.. Started experimenting a _lot_ with font hinting. The trouble there? There are a ton of different ways to do this, and doing it properly can occasionally mean letting go of subpixel positioning, unless you do some other various things like grabbing the advance width from `f->face->glyph->linearHoriAdvance` instead of `f->face->glyph->advance.x`, and of course then shifting over by 16 bits instead of 6 bits. Which FreeType has a handy function available for doing, `FT_MulFix()`. But if you try to use overly-sharpened grid fitting for the hinting (this is done by ORing the glyph's load flags with `FT_LOAD_TARGET_MONO`, but setting the render mode to `FT_RENDER_MODE_LCD` or `..._LIGHT`), this backfires tremendously since all but the first glyph will be very blurry. If we then backtrack and use `advance.x` again, there are _still_ problems: glyphs that come after a glyph that was kerned will be very blurry, while glyphs before the kerned glyph will still be sharp.
> 
> This led me to discovering an odd problem that I'm not sure anyone here has noticed yet:
> 
> Many fonts are not properly kerned, and this is _not_ Gargoyle's fault at all. FreeType does not support the GPOS table, which most OpenType fonts use to specify kerning, and an increasing number of TrueType fonts are using as well for kerning... Including the copy of Noto Serif that is currently included in FreeType. This means that a default copy of Gargoyle does not have kerning.
> 
> It turns out that the GPOS table can contain very advanced kerning instructions that allow kerning to depend on multiple glyphs surrounding the kerned glyph in question, so that contextual kerning can be done intelligently. Since FreeType only supports kerning of sets of kerned glyph pairs, GPOS is considered out of scope. Instead, Harfbuzz is recommended - and it's a library that can also support things like a font's internal custom ligatures and whatnot. Basically it does a lot of things Gargoyle currently does itself, and probably does a better job. We might want to use it.

I'd be fine with switching to Harfbuzz if it gives improved output.  It's available on Linux and Mac and while I don't have a copy built for MinGW, it shouldn't be too difficult to build one.  It'll increase the number of dependencies quite a bit, but I don't think that matters at all.

> Yesterday I was recovering from my second vaccination shot, so I apologize for not seeing your post until today. But overall I've just been kinda feeling overwhelmed (I know, entirely my own fault for taking on a bunch of ideas simultaneously). I'm not really sure what, if anything, should be changed.

No problem!  I just didn't want you to get a nasty surprise if you had something that got clobbered by the changes.  They're not _massively_ invasive but still would probably require some touch-ups.  The general structure of the code is the same, at least.

> Oh yeah, I also started messing with the initialization of `gli_image_rgb` to use 3 RGB subpixels instead of 4, using `QImage::Format_RGB888` so that the underlying pixel format matches what FreeType outputs. I got all that working locally, and the next step there is to rewrite the `draw_pixel` functions to just operate on one subpixel at a time, letting the `draw_bitmap` functions loop over all the subpixels instead of all the pixels.
> 
> However, while this would greatly simplify writing and maintaining these functions... I'm not sure I can really advocate for this. I'd kinda like to see that one feature of Windows Git/Glulxe that lets Narcolepsy define a transparent area of the window be implemented, but for that to happen we'd need that fourth subpixel so we can tell Qt which parts of the window to make transparent. I've seen semi-transparent sections of windows in Qt applications before, so I _know_ it's possible.

If the pixel format ever does change, the only thing that would need to be tested is macOS (and this would be one reason to switch macOS to Qt as well: consistency!)  I don't know about transparency in Qt, especially given how Gargoyle is using it (basically as nothing more than a canvas to draw on), but at some point I can look into it.  I've never seen Narcolepsy in Windows Git or Glulxe so I'll have to give that a look to see the effect you're talking about.

OK then, I'll get this merged in now, so there's less pain for future work.

Thanks again for all the work you're putting in to improve Gargoyle!

Issue #470, opened 2021-07-06
Restoring does not display files with no extension


jusw85:
On Linux, not tested on Windows.

In the older GTK version, when I `save` and specify a filename such as `a`, it would save as that filename `a`. `restore` would display that file in the file picker.

In the newer QT verison, when I `save` and specify a filename such as `a`, it would save as that filename `a`, without any extension (similar to GTK). However, `restore` would only filter files with `.sav` extension in the file picker. I'm not able to files with other / no extension. I have to rename `a` to `a.sav` in order to view it in the file picker for `restore`.

cspiegel:
First, I'm unable to reproduce the behavior you're describing in the earlier version.

For me, when the Gtk2 (or Gtk3) file dialog pops up, it has `Untitled.sav` as a default, with `Untitled` selected, so if I type a name, it already has `.sav` appended to it.  I can, of course, overwrite the whole name including extension if I wish; is that what you're referring to?

If I do that and save the file as `a`, it saves the file without an extension. But if I restore, I do not see `a`.  The restore dialog, for me, shows just `.sav` files.  I'm not running under a gtk-based desktop, though, so I don't know if that plays a role.

I don't think showing unsuffixed files is a viable answer (it's not possible anyway, in any useful fashion: the filter would have to be `*` which matches all files).  I have a pull request (which I don't intend to merge) at #473 which has a couple of approaches:

1. Add "All files (*)" as a filter in addition to the extension-specific filter; at least this gives easy access to files
2. Call `QFileDialog::setDefaultSuffix` to set the default suffix to `.sav` (or whatever is appropriate)

I think I'll actually wind up implementing 1, but 2 has a probably-fatal flaw, though you can give the branch a try.  The flaw is that, if gtk3 is used as a platform theme (e.g. `QT_QPA_PLATFORMTHEME=gtk3 gargoyle zork1.z3`, which is the default when running in a Gtk desktop), the Gtk3 file picker apparently has no way of adding a default extension.  What happens is the user selects a file (e.g. `a`), and then Qt appends `.sav` to it.  If `a.sav` exists, but `a` doesn't, the file dialog won't warn the user and ask about overwriting, but then since `.sav` is appended later, the file _will_ get overwritten.  `setDefaultSuffix` works for all other platforms I've tried (Qt, KDE, qt5ct, Haiku), but I think the lack of Gtk support means it's just not viable.

[...]

PR #473, opened 2021-07-07
Set a default suffix when saving files

cspiegel:


curiousdannii:
Can garglk use the recommended Glk extensions of .glksave, .glkdata, and .txt?

If the open dialogs can do the common pattern of having a filter for ".glksave" and then an option for all files that would be the best way I think of letting you still access old save files.

cspiegel:
> Can garglk use the recommended Glk extensions of .glksave, .glkdata, and .txt?
> 
> If the open dialogs can do the common pattern of having a filter for ".glksave" and then an option for all files that would be the best way I think of letting you still access old save files. (It might already be doing that, I haven't checked.)

I think that's reasonable.  This particular pull request here won't be merged in (See #470 for a hopefully-somewhat-clear description of what's going on), but once #470 is done in a different way, I'll create a new pull request for the updated extensions.

cspiegel:
Closing in favor of #474.

PR #499, opened 2021-08-09
Fix several warnings in interpreters

cspiegel:
These are all warnings which indicate invalid code (mismatched
pointers). A compiler would be perfectly justified in refusing to
compile this code, given that these are constraint violations rather
than potential undefined behavior, or style issues.

Clang still raises several warnings, most of which are purely related to
style, although a few are more worrisome, while still being valid code.
Those will be examined in the future: the present commit is solely to
avoid constraint violations.

### DIFF
--- a/terps/agility/os_glk.c
+++ b/terps/agility/os_glk.c
@@ -812,7 +812,7 @@ agt_statline (const char *cp_string)
 
   free (gagt_status_buffer);
   gagt_status_buffer = gagt_malloc (strlen (cp_string) + 1);
-  gagt_cp_to_iso (cp_string, gagt_status_buffer);
+  gagt_cp_to_iso ((const unsigned char *)cp_string, (unsigned char *)gagt_status_buffer);
 
   gagt_debug ("agt_statline", "string='%s'", cp_string);
 }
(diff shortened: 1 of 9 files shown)


angstsmurf:
Have you tried using the Clang Static Analyzer on the interpreters as well? I'd like to hear your opinion on the hundreds of things it finds in the Tads source.

cspiegel:
I have, and the results were demoralizing...  I just ran it through again as a refresher and it's about as bad as I remembered.

The biggest problem is that so many of these issues require a good understanding of the code, and unless it's Bocfel, I just don't know enough to do any real work on the interpreters in this area.  It'd take so much time (especially in TADS) to understand the codebase that it's basically impossible.  The one issue in Bocfel is a false positive, by the way: I've worked through the report in the past and determined that it's not actually possible for the pointer to be null.

I suspect a decent chunk of these issues are corner cases that are rarely, if ever, triggered.  One useful thing might be to run the interpreters through Valgrind and/or AddressSanitizer, to get an idea of what bugs are actually triggered under normal use.  I've done this for garglk and Bocfel on several occasions, but never the other interpreters.

angstsmurf:
Yeah, it seems to me that with every new release of Xcode, the analyzer finds a hundred or so new serious-sounding problems in the TADS source, but I don't think I've ever seen a TADS game actually crash. Keep up the good work!

PR #504, opened 2021-08-11
Remove bundled fonts

cspiegel:
This is a proposed update to how Gargoyle handles fonts.

The short of it is that bundled fonts are removed. The reason for this is licensing: the SIL Open Font License is incompatible with the GPL, so any fonts under that license can't be bundled.  That's the reason that Noto Serif replaced Bitstream Charter several years ago.  Charis SIL should have been the replacement font, but since it's under the OFL, it wasn't able to be.

Given that Charis SIL is based on Bitstream Charter, I'd like for it to be the "default" font, restoring the original look of Gargoyle.

Here are the salient points:

* Go Mono is still the fixed-width font, but Charis SIL is now the proportional font
* The fonts are included with Gargoyle under the names "Gargoyle Serif" and "Gargoyle Mono"; the fonts are identical to Charis SIL and Go Mono other than the names
  * This was done so that the fonts can be installed system-wide without clashing with any other fonts (e.g. if the user already has Charis SIL installled).  The Windows installer now installs these fonts, and the Unix install puts them in `/usr/share/fonts/gargoyle` which is typically indexed by fontconfig
* With no builtin fonts, it is possible for Gargoyle to fail to start due to missing fonts.  This is mitigated by employing the following search strategy for fonts:
  * If the user supplies a font name (e.g. propfont/monofont), and that font exists, it is used
  * Otherwise, a system-specific location is checked for the Gargoyle fonts (the install folder on Windows, or the location the fonts are installed on Unix, e.g. `/usr/share/fonts/gargoye`), and those fonts are used if found
  * Otherwise, if the Gargoyle fonts exist in a platform-specific location (which currently is defined as "the same directory as the gargoyle launcher", except on macOS, where nothing is used), they're used
  * Otherwise, the current directory is checked

This means that if you want to distribute Gargoyle standalone, you must include the Gargoyle fonts with it to guarantee success, but the lookup method should find them just fine.

I realize there's not much difference between "system-specific" and "platform-specific" as far as terminology goes.  "Platform-specific" maybe should unconditionally be "location of the executable".  It's really there to support AppImage on Linux, which is why I'm not concerned about implementing it on macOS.

From my testing, Linux AppImages work, macOS disk images work, and Windows installation works.


cspiegel:
@Tynach I know you had some thoughts about font selection, so you can take a look if you'd like.

curiousdannii:
Seems like installing fonts through the OS could be a minor hassle. (And would it make portable/no-install zips more difficult?) Seeing it will support looking for fonts in the same folder as the executable, wouldn't it be simpler to just put the fonts there? Then it would only need to look in OS locations for user-specified custom fonts.

cspiegel:
> Seems like installing fonts through the OS could be a minor hassle. (And would it make portable/no-install zips more difficult?)
> 
> Seeing as it will support looking for fonts in the same folder as the executable, wouldn't it be simpler to just put the fonts there? Then it would only need to look in OS locations for user-specified custom fonts.

For Unix systems, the executable will typically go in `/usr/bin` or similar, and it's not appropriate to put fonts in such a directory.  But the way this code handles that is to have a set font directory (`/usr/share/fonts/gargoyle`, for example), and have Gargoyle look there for the fonts. For portable installs, the executable directory is also used as a fallback, so that it will work whether or not the fonts are installed system-wide.  Basically, it casts a wide net in order to try to catch all the use-cases.

Having the fonts installed system-wide is just an option, and one that comes "free" due to the fact that Gargoyle has to look at system fonts, since users are allowed to select any font on the system.  The fallbacks are there so that the fonts don't need to be installed system-wide.

curiousdannii:
Ah, sure, that makes sense. Sounds like you're covering all the bases.

cspiegel:
> Ah, sure, that makes sense. Sounds like you're covering all the bases.

I hope so, anyway!  Bundled fonts are nice because you never even have to think about them; but the license issue really is a sticking point.

I do think this covers the main use-cases with the slight annoyance that if you are distributing a standalone version, you have to include the font files. This is somewhat mitigated on Linux with the AppImage (which contains the fonts already), and possibly on macOS (which is similar to AppImages in that the .dmg files contain the fonts).  The lone difficulty is Windows, but even then it's not too hard to ensure the fonts are in the same folder.  An annoyance, yes, but probably an acceptable one, since it's easy to do.

Tynach:
> @Tynach I know you had some thoughts about font selection, so you can take a look if you'd like.

I know it's a bit late, @cspiegel, but I just wanted to add that there are complications I ran into with font selection. Turns out that many MANY fonts, including pretty much all of them I wanted to suggest, only include kerning information in OpenType's format.. Not in the traditional kerning tables that FreeType supports. This would mean requiring the use of something like HarfBuzz to stay true to the idea of applying good typography out of the box, and that would require a lot of code modifications.

It's been a while now since I really looked at it, and I might be able to learn at this point... But at the time I looked into trying to modify Gargoyle to use Harfbuzz, I had a hard time even understanding the terminology used in Harfbuzz's documentation. This led me down a rabbit hole of studying OpenType's font kerning, which led to a LOT more than just kerning, and I was messing around with creating my own fonts just to know what was being talked about. I still feel like I've barely scratched the surface, too.

The thing is, these fonts can have more than just kerning going on. They can have character spacings that depend on *context* (spacing things apart based on more than just kerning data), custom ligatures, and 'ligatures' that depend on all sorts of criteria that's much more complex than just 'two characters next to each other'.

Gargoyle's implementation of ligatures just deals with a limited set number of them... And the way it works doesn't even detect these new fonts and how it's implemented in them. So a newer font that has, say, the Gargoyle-supported `fi` ligature, will *not* display the ligature when 'fi' is rendered with Gargoyle.

Because of all this, Harfbuzz would be *replacing* an extensive number of bits within Gargoyle's font handling, and it wouldn't just be about kerning anymore.

Personally, I think it'd be VERY worth it, especially from the perspective of supporting beautiful typography...

... But it also has significant downsides when paired with some of the stuff I said in issue #508. Harfbuzz does not actually *render* the glyphs, it just integrates with FreeType and lets you make better decisions when choosing which glyph to render and where to render it. As a result, it cannot know about how the spacings are modified as a result of hinting.
___

Aand that's why I sorta ignored my notifications from this project for a while. I kept investigating both of these paths, found out they contradict each other and you really kinda have to choose one direction or the other, and that kinda led me to feeling like I couldn't adequately answer what to do.

I've thought of a few theoretical ideas for how to make them 'compatible' with each other, since FreeType *does* offer ways of telling how much the spacing of a glyph changes as a result of hinting. I figure maybe that offset can be applied to Harfbuzz's output? But Pango (used by GTK) wrestled with this as well recently when they switched over to using Harfbuzz, and.. They decided too simply drop ALL support for both monochrome rendering *and* bitmap fonts. Yes, Pango/GTK completely dropped support for bitmap font rendering over this very issue.

If they couldn't reconcile the differences, I doubt I could. Their efforts resulted in very funky-looking glyph spacing that just.. Did *not* look right. I haven't looked into exactly what they've tried, but given they *did* try (and I read about it elsewhere in forum threads where people were complaining about the changes and mentioning attempts to fix the issues), I figure they probably tried stuff that's similar to what I've thought of.

Tynach:
It's worth noting that these issues only affect horizontal hinting of glyphs, as that's what's usually affected by kerning. As long as horizontal hinting isn't used (so `FT_LOAD_TARGET_MONO` isn't used), hinting can be used with Harfbuzz just fine. Also, bitmap fonts aren't actually hinted, per se.. I don't exactly understand why Pango dropped support for bitmap fonts entirely, instead of just not supporting monochromatic rendering of outline fonts. Maybe because they wanted to drop the use of `FT_RENDER_MODE_MONO` itself? Maybe that helped simplify their codebase in a way they felt was worth it?

I realize that isn't really relevant *here*. But since it ties into why this got really complicated for me to research, I figure I may as well give the additional context.

Tynach:
@khaledhosny, that doesn't make sense. There's nothing to use on Linux *except* FreeType for rendering fonts. ~~Also, everything I'm reading online says that Pango used to get kerning data from FreeType, but now has switched to getting kerning data from Harfbuzz.~~

Aand I'm glad I did a bit more research before actually sending this posts through, as I found the changelog for Pango and found out what was actually happening.

Pango used to use FreeType for font *loading*, that is actually opening the font files and parsing them.. But now it's no longer doing that, instead only using FreeType for font *rendering*.

PR #532, opened 2021-10-29
Madd's modernizations, including auto loops and LP64 support.

MaddTheSane:


terps/geas/GeasRunner.hh
```diff
@@ -109,49 +109,49 @@ protected:
...
-  virtual GeasResult play_sound (std::string filename, bool looped, bool sync)
+  virtual GeasResult play_sound (const std::string &filename, bool looped, bool sync)
   { return r_not_supported; }
-
+  
   /* Asks the user to type a free format string
    */
   virtual std::string get_string () = 0;
-
+
```
cspiegel, line 144: A couple lines here had extraneous whitespace added; a very minor point, but I think worth retaining the original "just a newline".

cspiegel:
There's a [build failure](https://github.com/MaddTheSane/garglk/runs/4050620897?check_suite_focus=true) with MinGW at https://github.com/MaddTheSane/garglk/blob/geasFix/terps/geas/readfile.cc#L683 due to `size_t` and `uint` being equivalent on MinGW.

Maybe the easiest solution is to have a single `string_int()` that takes `std::intmax_t`?

MaddTheSane:
Oh yeah, I forgot there's still 32-bit platforms.

And Windows 64-bit isn't LP64.

cspiegel:
One last concern with `string_int()`: since you're using `std::uintmax_t`, negative values will be interpreted as large positive value.  It looks like at least one instance can pass a negative integer value: `geas_implementation::get_ivar()` can return -32768 in case of error, and that return value can be passed to `string_int()`.  I'm not sure how much that matters: I don't know where these values are getting sent as strings, ultimately, so it might be perfectly acceptable to do this.

I'll merge once I get your feedback on this issue.  Thanks for dragging geas into the 21st century!

MaddTheSane:
I think that if there's a call to `string_int` that isn't an `int`, `uint`, or `std::uintmax_t`, it would throw an error about `call to 'string_int' is ambiguous`. At least it does so on Clang.

cspiegel:
> I think that if there's a call to `string_int` that isn't an `int`, `uint`, or `std::uintmax_t`, it would throw an error about `call to 'string_int' is ambiguous`. At least it does so on Clang.

The issue is that a negative int will be converted to a large unsigned value. So -32768 will be written as some huge number instead of-32768.

PR #551, opened 2021-11-08
Update to Glk 0.7.5

cspiegel:
This incorporates changes from cheapglk as well as the latest gi_dispa
and gi_blorb from https://www.eblong.com/zarf/glk/. The cheapglk version
is current master due to a couple of bug fixes introduced since the
latest tagged version (1.0.6).


curiousdannii:
It makes quite a bit of sense to have the unchanged files in a folder, but something like cggestalt.c is thoroughly modified.

garglk/gi_dispa.c
```diff
@@ -71,8 +66,8 @@ static gidispatch_intconst_t intconstant_table[] = {
     { "gestalt_CharOutput_ExactPrint", (2) },
     { "gestalt_DateTime", (20) },
     { "gestalt_DrawImage", (7) },
-    { "gestalt_GarglkText", (0x1100) },
```
curiousdannii, line 74: The garglk text stuff shouldn't have been removed.

cspiegel:
> It makes quite a bit of sense to have the unchanged files in a folder, but something like cggestalt.c is thoroughly modified.

Unfortunately most of them are changed to some degree or other.  I did leave out `cgstyle.c` (now renamed to `style.c`) due to the fact that it's completely different, but even with `cggestalt.c`, it's useful to be able to compare any potential future changes so nothing new is missed.  I do admit that I had similar thoughts to you as I went through adding all those `#ifdef`s, though...

To a large degree it's theoretical because `cggestal.c` won't be changing much, and when/if it does, it's easy enough to do a git log to see why, but I created the cheapglk directory as a way to note which files were imported, even if modified.  Since Gargoyle's `cggestal.c` does descend from cheapglk's, I'd prefer to keep it there so I don't lose track of it in the future.

I'd absolutely love a Glk library which implements the common parts with hooks for implementation-specific functionality, but right now I've just got to pretend cheapglk does that, and this helps with that illusion.

garglk/gi_dispa.c
cspiegel: Yikes, good catch.  I just copied those in rather than doing a diff, completely forgetting the custom Gargoyle stuff there.

curiousdannii:
> I created the cheapglk directory as a way to note which files were imported, even if modified.

Ah, I understand. I wonder then if you made the PR have two commits, one the unmodified cheapglk files, and then all the garglk changes on top, if that would then make it easier to merge updates in the future, as you could use git merges?

cspiegel:
> > I created the cheapglk directory as a way to note which files were imported, even if modified.
> 
> Ah, I understand. I wonder then if you made the PR have two commits, one the unmodified cheapglk files, and then all the garglk changes on top, if that would then make it easier to merge updates in the future, as you could use git merges?

I thought about that and couldn't figure out a clean way to do it.  It would be, without a doubt, the best solution.  Before I do any merging of this, I'll spend some more time on the problem.  I probably abandoned it too quickly, given how cleanly it would aid in future updates.

curiousdannii:
I don't think orphan branches should be necessary. I'm not certain this would work, but here's how I'd do it. The good thing with Git is that you can just give it a try, and if it doesn't work, it doesn't matter, just delete the branches and go back.

From master branch, make a cheapglk branch, and add the unmodified files in the right places.
Then rebase this branch on top of the cheapglk branch. Hopefully it is smart enough to turn the added files into diffs on the now existing files.

The super messy files will be the ones you've converted to C++. I don't think there's going to be a good solution there, as rebase probably won't pick it up. Perhaps you could manually rename it in a commit first? In the future, if you try updating that file again, I think it will still be more messy than the ones you left as .c though.

cspiegel:
> From master branch, make a cheapglk branch, and add the unmodified files in the right places. Then rebase this branch on top of the cheapglk branch. Hopefully it is smart enough to turn the added files into diffs on the now existing files.

The issue here is that this will include all of Gargoyle in the cheapglk branch; and it'll be frozen in time, since master can't be re-merged into it (or else the upstream cheapglk files will be modified to have Gargoyle's changes).  It wouldn't be impossible, but I think it'd be less-than-ideal to have this frozen Gargoyle version which can never be removed: if it were removed, the changes would be propagated back into master and Gargoyle would disappear.

I can't rebase master onto anything because messing with history could really screw up anybody who has branched/forked this repo.  But merging is fine.

> The super messy files will be the ones you've converted to C++.

This one is a lot simpler than it looks at first glance, because Git can detect that the C file has been renamed to C++, and the merges are transparent.  My testing has indicated no problems with this.

> If you can't work it out, don't worry. The future changes aren't going to be that frequent. And you could always make a patch file from the cheapglk repository and then apply the patch file here. Patch files still have their uses!

I'm really hoping to avoid manual patching because the tooling associated with Git merge conflicts is good, and something familiar to me.  Interactively fixing a conflict with vim-conflicted is much nicer than manually handling reject files from patch.

But to make a long story short-ish, the orphan branch is analogous to making a branch off of master, but without the surrounding Gargoyle stuff that will forever be stuck in time.  I think it makes the intent more clear, since the branch contains _only_ the upstream cheapglk.

curiousdannii:
Alright, sounds like you're on top of it, and that your git experience is greater than mine anyway!

(And I wasn't suggesting rebasing master, just the glk-0.7.5 branch. But it doesn't matter.)

cspiegel:
Thanks for taking the time to look through/talk through this branch.  It's really helpful to have a second set of eyes on things, as well as a dissenting voice. I know I'm not always right, but when I am wrong, I don't know it unless somebody points it out!  This goes for all the various bug reports, comments, reviews, and so on that you've done over the years.

Issue #180, opened 2015-03-22
SCARE: Commands abbreviated twice


GoogleCodeExporter:
```
Commands like "examine", "look", "inventory", "wait" etc are abbreviated both 
in SCARE itself and in the code that connects SCARE to Glk. The result can be 
seen in the following transcript:

---

> z
[z -> wait]
Time passes...

> 'z
Time passes...

> glk abbreviations off

Glk abbreviation expansions are now off.

> z
Time passes...

---
The abbreviation handling from os_glk.c has the highest priority and it prints 
a message when it expands an abbreviation: [z -> wait]. This abbreviation 
handling can be bypassed by starting the input with a quote or can be turned 
off via a setting.

The abbreviation handilng from scrunner.c treats the one-letter command as an 
equal version of the longer command. For example "i", "inv" and "inventory" are 
treated equally. There is no expansion step and there is no way to turn it off.

I think that having two mechanisms for accepting short commands is unnecessary. 
It can also be confusing: for example in the transcript abbreviations are 
turned off but the short version is still accepted. And in some games the short 
versions of commands can conflict with custom commands from the game, having 
the ability to bypass abbreviations is useful then.

One way to address this would be to remove the Glk abbreviations. The advantage 
of this option is that it doesn't require any changes in SCARE.

Another way to address this would be to remove the short aliases from SCARE. 
The advantage of this option is that the user would gain the ability to bypass 
or turn off abbreviations.

I'm willing to write a patch for this, but I'd first like to hear the 
maintainer's opinion on which approach is preferred.

```

Original issue reported on code.google.com by `mthuu...@gmail.com` on 7 Feb 2012 at 10:16

PR #552, opened 2021-11-08
Don't expand abbreviation in the SCARE interpreter

cspiegel:
The Glk layer expands as well, so these are unnecessary, and according
to #180, this can cause problems with a specific (though unnamed)
game. Since Glk abbreviations can be turned off, this gives the most
flexibility.


curiousdannii:
For many of the unmaintained interpreters, I (and maybe some other people) have been using garglk for convenience. If I understand correctly, this would mean non-glk users of this repo's scare terp wouldn't have any shortcut abbreviations? Maybe we could put them inside a compile-time definition check instead of just removing them?

cspiegel:
> For many of the unmaintained interpreters, I (and maybe some other people) have been using garglk for convenience. If I understand correctly, this would mean non-glk users of this repo's scare terp wouldn't have any shortcut abbreviations? Maybe we could put them inside a compile-time definition check instead of just removing them?

That's a really good point.  I'll add such a check.

curiousdannii:
Beautiful, thank you!

Issue #409, opened 2021-02-27
arm64: tadsr starts but doesn't play any game


cibersheep:
I'm building the terps with remglk for Linux armhf, amd64 and arm64 architectures.
tadsr builds for all three architectures and works on armhf and amd64 but loading a game on arm64 after init, it will give this error:

`[TADS-609: unknown resource type in .gam file]` with these games
-[3.gam](https://ifdb.org/viewgame?id=1z2lxiqua980sedk)
-[museum.gam](https://ifdb.org/viewgame?id=7if80tf1uuu903y3)

`Runtime error: object value required`
-[comp02tr.t3](http://www.ifarchive.org/if-archive/games/tads/comp02tr.t3)

Those same games work on armhf and amd64,  build from same source.

Any clue what that might be?

cibersheep:
Running `file tadsr` on arm64 outputs:
```
tadsr: ELF 64-bit LSB executable, ARM aarch64, version 1 (GNU/Linux), 
dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, 
BuildID[sha1]=920592a8e104724f236b91f86d2f2032a2587897, not stripped
```

on amd64:
```
tadsr: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), 
dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, 
BuildID[sha1]=9eaf1e7bbd0debd84b08383349475a8b37992637, not stripped
```

on armhf:
```
tadsr: ELF 32-bit LSB executable, ARM, EABI5 version 1 (GNU/Linux), 
dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0,
BuildID[sha1]=636e2d4629d0fb569172dd81100daa8c83c2006c, not stripped
```

cspiegel:
No idea whatsoever, but you can maybe get a rough idea by hacking `fio.c` to display the value of `buf` when `ERR_UNKRSC` is raised (it looks like that triggers the "unknown resource type in .gam file" message).

In short, TADS has a whitelist of resource types it's expecting in `buf`, and someting isn't matching up.  At least the contents of `buf` on error might give _some_ indication of what's going on (short read?  bytes swapped?  etc)

cibersheep:
Thanks for the hint. I'll see if I can come up with some useful information

cibersheep:
[...]

PR #558, opened 2021-11-11
Fix TADS on ARM64

angstsmurf:
This is a very simple fix for #409. It makes TADS use the x86_64 header when building on ARM64.

There is a less hacky fix described [here](https://intfiction.org/t/please-test-spatterlight-on-apple-silicon/52476/9) which replaces the x86 headers with a platform-agnostic header, but I had a really hard time making it work on Spatterlight. In the end I got it to work by setting the TADS targets to build as C++11, but I never understood why it was needed.

I tried implementing this in Gargoyle [here](https://github.com/angstsmurf/garglk/tree/tads), but couldn't make it build.

Also, the fact that the TADS source files use Windows line endings makes them a pain to edit, but hopefully I didn't change any to the wrong format.

### DIFF
--- a/terps/tads/tads2/os.h
+++ b/terps/tads/tads2/os.h
@@ -75,7 +75,7 @@ extern "C" {
 #ifdef __ppc__
 #define _M_PPC
 #else
-#ifdef __x86_64__
+#if defined __x86_64__ || defined __aarch64__
 #define _M_IX86_64
 #else
 #ifndef _M_IX86


curiousdannii:
If this works you should suggest it in the forum too.

cspiegel:
> There is a less hacky fix described here which replaces the x86 headers with a platform-agnostic header, but I had a really hard time making it work on Spatterlight. In the end I got it to work by setting the TADS targets to build as C++11, but I never understood why it was needed.

> I tried implementing this in Gargoyle here, but couldn't make it build.

When I switched to C++11 for Gargoyle, it built fine.  This can be done with a patch like:

```
diff --git a/terps/tads/CMakeLists.txt b/terps/tads/CMakeLists.txt
index c8ac26d..b74264e 100644
--- a/terps/tads/CMakeLists.txt
+++ b/terps/tads/CMakeLists.txt
@@ -52,6 +52,7 @@ if(WITH_TADS)
     terp(tadsr
         SRCS ${srcs}
         MACROS ${macros}
+        CXXSTD 11
         POSIX)
 
     if("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang")
```
If that works for you as well, would your tads branch be a better solution to this problem?

> Also, the fact that the TADS source files use Windows line endings makes them a pain to edit, but hopefully I didn't change any to the wrong format.

I've been meaning to explore how easy it would be to use FrobTADS as the base for Gargoyle's TADS, since FrobTADS (and QTads) seem to be where active development on TADS interpreters is today.  I point that out because FrobTADS uses Unix line endings, so that'd just be a nice bonus.

angstsmurf:
Right, I set it to  C++11 and made another pull request with the alternative approach. It is a matter of taste which one is the best, really.

cspiegel:
Closing because #560 has been merged.

PR #560, opened 2021-11-12
Fix TADS on ARM64, alternative approach

angstsmurf:
This is an alternative fix for  #409. It is more invasive than #558, but replaces several header files with a single one. There is potential for future cleanup and simplification in [os.h](https://github.com/garglk/garglk/blob/master/terps/tads/tads2/os.h).


cspiegel:
As long as this approach works just as well as #558, I think it should be preferred.  FrobTADS is modern TADS, so if it's how they do it, it's how we ought to do it as well.

curiousdannii:
Doesn't look like FrobTads has the GPL licence.

It would be nice for the community though if we could coalesce around one de facto upstream repository for TADS. Like we did for Hugo.

cspiegel:
> Doesn't look like FrobTads has the GPL licence.

You're right, and that throws a huge monkey wrench into TADS generally, because upstream TADS isn't GPL either.  It has a license which looks very much in conflict with the GPL.  Excerpts:

> 2. You may not require or collect a fee for copies of this software, or any part of this software, that you give to other people.

The GPL allows a distribution fee.

> 4. You may not modify this software except as permitted below (see "derivative works")
> ...
> Derivative Works
> This source code is distributed for the specific purpose of porting TADS, so that you can run the software on any system of your choosing. All other derivative works are prohibited without the written permission of the author.

The GPL doesn't allow such a restriction: there is no prohibition of any derivative works, so long as they're under the GPL.

> 5. Anyone to whom you give a copy of this software receives all of the same permissions that you did under this license and is subject to all of the same restrictions.
> 6. You are not allowed to create derivative works, which are works that contain or are based on all or part of this work, except under the conditions described below.

Similar to the above, the GPL doesn't allow you to restrict others like this.

In short, unless I'm missing something, TADS can't be included with Gargoyle.  Gargoyle is GPL (*not* LGPL), so linking libgarglk requires GPL compatibility.

Or, at least, you can't distribute such a binary.  Glk TADS is legal (since you could create a Glk implementation which conforms to the TADS license), but once it's linked to a GPL Glk library, it's no longer legally distributable.

A conversion of Gargoyle to the LGPL is a solution, but would require the consent of all copyright holders, i.e. anybody who has modified the code.  I'm not sure how feasible even _asking_ everybody would be, and if anybody objects, we'd have to rewrite that part of the code (and if a major contributor objects, rewriting wouldn't even be realistic).

curiousdannii:
I thought at some point in the last MJ Roberts gave permission for TADS to be included in Gargoyle under the GPL licence?

cspiegel:
> I thought at some point in the past MJ Roberts gave permission for TADS to be included in Gargoyle under the GPL licence?

I realized that was a possibility after posting, did a search, and came up with:

https://intfiction.org/t/upgrading-gargoyle-to-use-recent-tads-sources/9141/5
https://intfiction.org/t/tads3-if-only-it-were-open-source/41469/7

So you're absolutely right, and I swear I'm having a déjà vu here, as I feel I've gone through this exact reasoning before...

Anyway, crisis averted!

curiousdannii:
Sorry, edited my post as you were posting...

cspiegel:
No worries, we've come to exactly the same (good) conclusion.  Technically speaking it looks like QTads has a GPL TADS core while FrobTADS does not, so QTads is probably the right upstream source at the moment.  Once this is merged I'll see how it looks pulling from QTads.

Sorry for hijacking your pull request, @angstsmurf !

cspiegel:
I'll merge this one in and at some point look toward updating everything from QTads.  Thanks!

PR #567, opened 2021-11-14
Fix style setting

cspiegel:
This got missed in the update to cheapglk for Glk 0.7.5.

### DIFF
--- a/garglk/cheapglk/cgstream.c
+++ b/garglk/cheapglk/cgstream.c
@@ -21,7 +21,7 @@
  *                                                                            *
  *****************************************************************************/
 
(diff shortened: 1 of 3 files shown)


garglk/cheapglk/cgstream.c
```diff
@@ -1719,6 +1726,9 @@ void glk_set_style_stream(stream_t *str, glui32 val)
         return;
     }
     /* This cheap library doesn't handle styles */
+#ifdef GARGLK
+    gli_set_style(gli_currentstr, val);
```
curiousdannii, line 1730: Should be str not currentstr.

cspiegel: Indeed.  Thank you!

Issue #371, opened 2020-04-01
macOS: Sounds will keep playing after game window is closed


angstsmurf:
Closing the window of a game (but not the application) while sound is playing on macOS will not stop the sound.

To reproduce with [Soundtest](https://github.com/angstsmurf/soundtest), just type LOOP at the prompt and then close the window.

Quitting the Gargoyle application will stop any sounds as expected.

Other platforms do not support keeping the Gargoyle application open after closing the game window, or opening several games using the same Gargoyle instance, so they are unaffected by this.

cspiegel:
@angstsmurf Can you please test on master now that #545 has been merged in?  It's possible that this bug is related to that one, but I have macOS installed in a VM with no sound support, so I'm unable to check this myself.

angstsmurf:
I'm afraid this still happens. Is the SDL sound running in a separate process that is not killed by `kill(processID, SIGUSR2)` in [launchmac.mm](https://github.com/garglk/garglk/blob/master/garglk/launchmac.mm#L549-L556)?

I tried to stop the sounds by iterating through all Glk sound channels and calling `glk_schannel_stop()` on them in this method, but it had no effect.

angstsmurf:
I found the place to do it from inside the interpreter process. It should have been obvious. See #596.

PR #596, opened 2021-12-07
macOS: Stop sound channels when terminating game

angstsmurf:
This is a fix for #371. There might be a more elegant way to do it, but it seems to work.

When the interpreter process receives a termination signal, we stop any Glk sound channels.


cspiegel:
Great, thank you!

PR #603, opened 2022-01-01
Update TADS with code from new upstream

curiousdannii:
Building with CMake, it works when I run tadsr directly, but when running gargoyle it complains it can't find tadsr... not sure what's going wrong there, and if it's my fault or not. Could it just be that it can't find the binary if you don't install it?

Most of the changes are line ending changes. This hides whitespace only changes: https://github.com/garglk/garglk/pull/603/files?diff=unified&w=1


cspiegel:
> Building with CMake, it works when I run tadsr directly, but when running gargoyle it complains it can't find tadsr... not sure what's going wrong there, and if it's my fault or not. Could it just be that it can't find the binary if you don't install it?

Yes, that's correct.  For development you can set the environment variable `GARGLK_INTERPRETER_DIR` to the directory containing the interpreters; although I just realized that since TADS is in a subdirectory in the CMake build directory, you can either point the variable to TADS or to the other interpreters, but not all at once!  I'll probably pull the TADS binary out of its own subdirectory at some point just to make development slightly easier.

Looks good, and I wholeheartedly approve of the whitespace changes.  DOS line endings were only used because I didn't want to diverge from the original upstream, but with the new upstream having Unix line endings, the change is very welcome.

PR #607, opened 2022-01-28
Update Bocfel to 1.4.2

cspiegel:


curiousdannii:
Are you essentially going outside of Glk to allow for a filename prompt to be given? Would it be worth making a quick micro Glk extension so that other interpreters or Glk libraries could use the same API?

cspiegel:
> Are you essentially going outside of Glk to allow for a filename prompt to be given? Would it be worth making a quick micro Glk extension so that other interpreters or Glk libraries could use the same API?

No, as of right now this doesn't handle the case of "prompt with user-supplied filename".  If a prompt is requested, Bocfel is still at the mercy of whatever Glk implementation is being used.  The change here is just to allow the interpreter to specify a filename in non-prompt mode (e.g. SameGame will be able to silently read and write the file `SAMEGAME.HGH` to track high scores).

I've pondered adding an extension to support filename suggestions, but for this release I punted.  I may come back to it at some point.

PR #610, opened 2022-01-31
Unix: Install garglk.ini to `CMAKE_INSTALL_SYSCONFDIR`

akien-mga:
There might have been a good reason not to do so, so feel free to close this PR if it's not wanted.

Installing `garglk.ini` to `/etc/garglk.ini` seems to be the main path where Gargoyle expects to find it on Unix (at least on Linux), and what I used to do manually in my distro packaging of gargoyle (for Mageia).

I'm not sure how it works on macOS, nor whether it should also be done for the Linux AppImage distribution. I can amend the commit as needed.

### DIFF
--- a/garglk/CMakeLists.txt
+++ b/garglk/CMakeLists.txt
@@ -7,7 +7,10 @@ option(BUILD_SHARED_LIBS "Build a shared libgarglk instead of a static library" 
 
 if(UNIX AND NOT APPLE)
     option(WITH_FREEDESKTOP "Install freedesktop.org application, icon, and MIME files" ON)
-    set(GARGLKINI "/etc/garglk.ini" CACHE STRING "Full path for garglk.ini")
+    set(GARGLKINI "${CMAKE_INSTALL_FULL_SYSCONFDIR}/garglk.ini")
+    if(NOT APPIMAGE)
+        option(INSTALL_GARGLKINI "Install an example config file to ${GARGLKINI}" OFF)
+    endif()
 endif()
 
 if(APPLE)
@@ -219,6 +222,10 @@ elseif(UNIX)
     install(FILES cheapglk/gi_blorb.h cheapglk/glk.h glkstart.h DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/garglk")
     configure_file("garglk.pc.in" "${CMAKE_CURRENT_BINARY_DIR}/garglk.pc" @ONLY)
     install(FILES "${CMAKE_CURRENT_BINARY_DIR}/garglk.pc" DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
+
+    if(INSTALL_GARGLKINI)
+        install(FILES "garglk.ini" DESTINATION "${CMAKE_INSTALL_SYSCONFDIR}")
+    endif()
 
     if(WITH_FREEDESKTOP)
         install(FILES "gargoyle.desktop" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/applications")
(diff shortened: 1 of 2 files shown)


akien-mga:
Thanks for the detailed explanation, that makes a lot of sense!

I hadn't seen that Gargoyle can already generate a custom `garglk.ini` via Ctrl+Comma, but given that I agree that installing a default one in `/etc` is not necessarily what most users need. I think distro packagers install it out of habit ("If there's an option for a system-wide config, it must be needed/useful?"), but it does sound like a configuration option which is still nice to have but somewhat of a legacy feature.

So I implemented `INSTALL_GARGLKINI` as suggested, defaulting to OFF for now, but I'm happy to change it to ON if you think that's better.

garglk/garglk.ini
```diff
@@ -6,7 +6,7 @@
 # order, with later settings overriding earlier settings:
 #
 #   1: same directory as the executable: garglk.ini (windows)
-#   2: /etc/garglk.ini (unix)
+#   2: /etc/garglk.ini (or other location set at build time, Unix only)
```
akien-mga, line 9: Changed to match:
```
garglk/config.cpp
235:// 5. /etc/garglk.ini (or other location set at build time, Unix only)
```

garglk/CMakeLists.txt
```diff
@@ -220,6 +223,10 @@ elseif(UNIX)
     configure_file("garglk.pc.in" "${CMAKE_CURRENT_BINARY_DIR}/garglk.pc" @ONLY)
     install(FILES "${CMAKE_CURRENT_BINARY_DIR}/garglk.pc" DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
 
+    if(INSTALL_GARGLKINI)
+        install(FILES "garglk.ini" DESTINATION "${GARGLKINI}")
```
cspiegel, line 227: Instead of `${GARGLKINI}` this should be `${CMAKE_INSTALL_SYSCONFDIR}`.  Otherwise it will think the directory is called `/etc/garglk.ini` and install a file `/etc/garglk.ini/garglk.ini`.

But then there's a disconnect (and that disconnect is why you used `${GARGLKINI}`, I'm sure): `${GARGLKINI}` is guaranteed to have the right value, so it should be used directly for installing `garglk.ini`.

The only reason `GARGLKINI` can be specified is because the old Jam build system supported it, so I carried it over.  But there's no real reason to allow an arbitrary file as the default config.  So my (last!) request is this:

Always set `GARGLKINI` based on `${CMAKE_INSTALL_SYSCONFDIR}`: don't make it a user option.  This should be set unconditionally: whether or not the user wants to _install_ the config file, it should always know to look for it in the sysconf dir.

If `INSTALL_GARGLKINI` is set, then install to `${CMAKE_INSTALL_SYSCONFDIR}`.

If the user wants a different directory, `-DCMAKE_INSTALL_SYSCONFDIR` can be used to set it.  If the user wants a different name than `garglk.ini`, that's not going to be possible.

Thanks for working on this!

cspiegel:
Everything looks great, thank you!

autumnontape:
Just for the sake of explaining myself as the maintainer of the `gargoyle` package on the AUR, I install `/etc/garglk.ini` mainly because I think its presence makes it a little easier for someone familiar with Linux-style configuration files to figure out how to configure Gargoyle. If you find a configuration file in `/etc`, you expect to be able to override it with a corresponding file in `~/.config`. The way Pacman handles config file upgrades (if the user's modified the file, it appends `.pacnew` to the path it installs the new version to and notifies the user) can also be helpful when configuration formats change... I don't configure Gargoyle by editing the system-wide config file myself, but I can imagine someone wanting to handle things that way.

Issue #613, opened 2022-01-31
garglk.ini not taking effect on macOS


dfabulich:
In Gargoyle version 2022.1 on macOS Monterey 12.2, changes to garglk.ini aren't taking effect.

To repro:
1. Delete `~/garglk.ini`.
2. Launch Gargoyle and run a game. (I'm running Violet. <https://ifarchive.org/if-archive/games/zcode/Violet.zblorb>)
2. Go to the Gargoyle -> Preferences menu. That creates `~/garglk.ini`.
3. Set the `zoom` factor to `2.0`, save, and quit Gargoyle.
4. Launch Violet again.

Expected: The window, text, etc. should be twice as large
Actual: No effect

I also tried setting `graphics      0` to disable graphics, and running Counterfeit Monkey <https://github.com/i7/counterfeit-monkey/releases/download/r10/CounterfeitMonkey-10.gblorb> (type `yes` and then `andra` to see graphics). With graphics disabled, I expected the map not to be visible, but it was.

cspiegel:
Definite bug!  There is, at least, a painless workaround:

`ln -s $HOME/garglk.ini $HOME/.garglkrc`

Gargoyle uses the XDG Base Directory Specification for most operating systems (all but Windows and Haiku) as well as directly looking at `$HOME/.garglkrc` for backward compatibility.  As such you can just link the file that Gargoyle edits (`$HOME/garglk.ini`) to a location it actually looks at.

I think the right fix is to completely abandon XDG on macOS and use the `$HOME/Library` to store the configuration file.

cspiegel:
Pull request #619 addresses this; not by switching to `$HOME/Library` (yet) but by ensuring `$HOME/garglk.ini` is considered a config file.

PR #619, opened 2022-01-31
Ensure $HOME/garglk.ini is loaded on macOS

cspiegel:
See #613


gargoyle_osx.sh
```diff
@@ -132,6 +132,8 @@ cp fonts/Gargoyle*.ttf $BUNDLE/Resources/Fonts
 
 codesign -s - -f --deep Gargoyle.app
 
+exit
+
```
dfabulich, line 136: What’s this?

cspiegel: Accidental commit from testing; I realized I left it here, came to fix it, and saw you beat me by 10 minutes!

PR #621, opened 2022-02-01
Load libspeechd with dlopen instead of depending on it directly

autumnontape:
This makes it possible to build and run a TTS-enabled version of Gargoyle on a system without speech-dispatcher installed.

garglk will only attempt to load libspeechd.so.2 if speech is enabled in the configuration, and it will print a warning to stderr if it isn't found or if the version that was found is too old.


cspiegel:
I completely understand the desire for this from a package maintainer's standpoint.  Given that, I'm in favor of supporting the dynamic loading of speech-dispatcher.

However, I would like for it to be opt in, with the previous behavior the default.  Using dlopen() is more fragile: if the next speech-dispatcher release changes the ABI, the current implementation will work with a recompile.  The dlopen() method will require source changes to support the new ABI (even if it's just updating the name of the library).  So `WITH_TTS` will have to become a quad-state: ON, OFF, AUTO, and DLOPEN (or whatever).  That will allow packagers (who can guarantee the existence of a specific speech-dispatcher version) to offer a single Gargoyle package, while allowing others to build against whatever speech-dispatcher version they have at their disposal.

garglk/CMakeLists.txt
```diff
@@ -166,20 +166,7 @@ if(WITH_TTS)
...
-
-        pkg_check_modules(SPEECH_DISPATCHER ${TTS_REQUIRED} IMPORTED_TARGET speech-dispatcher)
-
-        if(SPEECH_DISPATCHER_FOUND)
-            target_sources(garglk PRIVATE ttsspeechd.cpp)
-            target_link_libraries(garglk PRIVATE PkgConfig::SPEECH_DISPATCHER)
-        else()
-            target_sources(garglk PRIVATE ttsnull.c)
-        endif()
+        target_sources(garglk PRIVATE ttsspeechd.cpp)
```
cspiegel, line 169: For dlopen(), you'll also need to link against `${CMAKE_DL_LIBS}` for platforms which require a specific library for support (e.g. `libdl`).

garglk/ttsspeechd.cpp
```diff
@@ -21,20 +21,106 @@
  *                                                                            *
  *****************************************************************************/
 
+#include <dlfcn.h>
+#include <stdio.h>
```
cspiegel, line 25: Please use `std::cerr` from `<iostream>` to output to standard error.

autumnontape:
@cspiegel I can make that change. You think that's necessary for speech-dispatcher? As I wrote in a comment, this implementation is following Firefox's example, and it doesn't look like speechd has had an ABI-breaking change in many years.

autumnontape:
I'd propose "DYNAMIC" for WITH_TTS value corresponding to this implementation.

cspiegel:
> I can make that change. You think that's necessary for speech-dispatcher? As I wrote in a comment, this implementation is following Firefox's example, and it doesn't look like speechd has had an ABI-breaking change in many years.

I would prefer that it still be able to link everything in the "normal" fashion, yes.  I consider using dlopen() to be a special case, since its necessity is one of convenience for packagers.  If you don't need a single binary that can optionally use speech-dispatcher, then there's no need for the dlopen() approach with its potential downside.  But providing the option to packagers is, I think, a pragmatic approach.

> I'd propose "DYNAMIC" for WITH_TTS value corresponding to this implementation.

That sounds good to me.

autumnontape:
@cspiegel All right, I've added a "DYNAMIC" value for WITH_TTS, and I've documented it in INSTALL.md as well. The default is still "AUTO."

cspiegel:
Looks good, thank you!

PR #627, opened 2022-02-17
Updated ScottFree

angstsmurf:
I don't know if there is any interest in this, but I think that with a bit of polish, this could eventually become a nice update to the ScottFree interpreter.

It is a cut-down version of the Spatterlight ScottFree fork, which removes all support for German grammar, graphics, and ZX Spectrum and Commodore 64 versions, although there are traces of them left in the code. The TI-99/4A format support is still in there, but not very well tested.

New features:
- Restart, undo, ram save, ram load, transcript.
- Universal abbreviations: n, s, e, w, u, d for directions both as verbs and nouns, i for inventory, x for examine, l for look, and z for wait.
- Improved TAKE ALL and DROP ALL which are supposed to work in all games without breaking stuff.
- Commands can be chained together by commas or periods.
- Multiple nouns such as TAKE AXE AND FISH work.
- Changes in room descriptions will create pauses where appropriate, so that they are visible to the player. There should probably be a way to switch this off.
- Unicode support for non-ASCII input and output.
- Line wrapping in the upper window.

EDIT: Now also supports IT (refers to the last noun) and TAKE ALL BUT X AND Y


terps/scott/detectgame.c
```diff
@@ -0,0 +1,190 @@
...
+        SkipList[i] = EnglishSkipList[i];
+    for (int i = 0; i < NUMBER_OF_DELIMITERS; i++)
+        DelimiterList[i] = EnglishDelimiterList[i];
+    for (int i = 0; i < NUMBER_OF_EXTRA_NOUNS; i++)
+        ExtraNouns[i] = EnglishExtraNouns[i];
+
+    // Check if the original ScottFree LoadDatabase() function can read the file.
+    if (LoadDatabase(f, 0)) {
+        fclose(f);
+        GameInfo = MemAlloc(sizeof(GameInfo));
```
cspiegel, line 144: This should be `sizeof(*GameInfo)`.

terps/scott/gameinfo.c
```diff
@@ -0,0 +1,188 @@
...
+//  Created by Administrator on 2022-01-30.
+//
+
+#include <stdio.h>
+
+#include "definitions.h"
+
+/* This is supposed to be the original ScottFree system
+ messages in second person, as far as possible */
+char *sysdict[MAX_SYSMESS] = {
```
cspiegel, line 14: Presumably this would be better as `const char *`.

cspiegel: Well, I know I'm being annoying here, but there are lots of places that ought to be const throughout the code. Anything which is just string literals should be stored in `const char*`, and when those changes are made, warnings will radiate out to various other functions/variables, but it's simple enough to fix as you find them.

The Glk API, unfortunately, takes some `char *` values when it should take `const char *`, so casts will be required there, but throughout ScottFree things ought to be const except where they're modified.

terps/scott/layouttext.c
```diff
@@ -0,0 +1,94 @@
...
+#include <string.h>
+
+#include "layouttext.h"
+
+int FindBreak(const char *buf, int pos, int columns)
+{
+
+    int diff = 0;
+
+    while (diff < columns && !isspace(buf[pos])) {
```
cspiegel, line 19: This is always one of my least-favorite review issues, because it seems so ridiculous, but when passing a `char` to the `ctype` functions, you need to cast it to `unsigned char`.  Why?  Because the C standard says so, that's why...

But it's really to accommodate systems which use macros plus lookup tables to speed up these functions.  Solaris used to do this, and to my recollection NetBSD did as well.  I'm not sure any modern systems still do, but they're allowed to.  On Solaris, `isspace` used to look like this:

```#define	isspace(c)	((__ctype + 1)[c] & _S)```

And `__ctype` is an array mapping `unsigned char` to a set of bits indicating which properties each character has.  The `+ 1` is because EOF (-1 on Solaris and probably ever other existing system) has to be handled as well.  So `isspace(x)` will index directly into the array, and if you have a signed `char` type, and the value is negative, you'll index backward into something that's totally unrelated.

So, long story short, all `ctype` functions are required to  take “an `int`, the value of which shall be representable as an `unsigned char` or shall equal the macro `EOF`.” That's backwards compatibility for you.

File: terps/scott/load_TI99_4a.c [hunk 4]
```diff
@@ -0,0 +1,816 @@
...
+        dict[i] = MemAlloc(word_len + 1);
+
+        strncpy((char *)dict[i], w1, word_len);
+
+        i += 1;
+    }
+}
+
+struct Keyword {
+    char *name;
```
cspiegel, line 243: Recommend making this a `const char *.`

terps/scott/load_TI99_4a.c
```diff
@@ -0,0 +1,816 @@
...
+    GameHeader.PlayerRoom = pr;
+    GameHeader.LightTime = lt;
+    LightRefill = lt;
+    GameHeader.NumMessages = mn;
+    Messages = MemAlloc(sizeof(char *) * (mn + 1));
+    GameHeader.TreasureRoom = trm;
+
+    int offset;
+
+#pragma mark rooms
```
cspiegel, line 683: When warnings are enabled with gcc (which they're not, for some reason, but should be: I must have just forgotten to add it), these pragmas cause warnings.  They ought to be wrapped in tests to see if the compiler is clang.

terps/scott/scott.c
```diff
@@ -33,1379 +33,2095 @@
...
+#ifdef SPATTERLIGHT
+extern glui32 gli_determinism;
 #endif
-static void Display(winid_t w, const char *fmt, ...)
+
+static const char *game_file;
+
+//#ifdef __GNUC__
+//__attribute__((__format__(__printf__, 2, 3)))
+//#endif
```
cspiegel: This can be removed

terps/scott/gameinfo.c
angstsmurf: Definitely not annoying! Thanks the review. I hope I got all of them.

cspiegel:
@angstsmurf Do you feel this is in a pretty good state for a full review again?  I'm happy to look it over and bring it in to Gargoyle if you're fine with its current state.

terps/scott/TI99_4a_terp.c
```diff
@@ -0,0 +1,570 @@
...
+
+#include "glk.h"
+#include "load_TI99_4a.h"
+#include "scott.h"
+#include "TI99_4a_terp.h"
+
+ActionResultType PerformTI99Line(uint8_t *action_line)
+{
+    if (action_line == NULL)
+        return 1;
```
cspiegel, line 11: Since this returns an enum, it ought to use the enum names (ACT_FAILURE here, preusmably).

terps/scott/TI99_4a_terp.c
```diff
@@ -0,0 +1,570 @@
+#include <string.h>
+
+#include "glk.h"
+#include "load_TI99_4a.h"
+#include "scott.h"
+#include "TI99_4a_terp.h"
+
+ActionResultType PerformTI99Line(uint8_t *action_line)
```
cspiegel, line 8: This should be `static` since it's only used in this file.

In addition, it looks like `action_line` can be `const`.

terps/scott/TI99_4a_terp.c
angstsmurf: Done!

terps/scott/TI99_4a_terp.c
angstsmurf: And done.

angstsmurf:
I don't think there is a lot to be done here except testing games, unless you have more comments. I suppose I could add some kind of script reading mechanism to make testing easier.

Also, the Mysterious Adventures (and perhaps others) need the PREHISTORIC_LAMP setting that destroys the lamp when it runs out, and there is no way to set this in Gargoyle. This problem exists in the current ScottFree in master as well. Perhaps I could just calculate checksums and autodetect these games.

angstsmurf:
Now the lamp should work correctly in the Mysterious Adventures.

I also implemented a way to read command script, (#SCRIPT READ) and have played through The Golden Baton, Waxworks, Secret Mission, Voodoo Castle, Strange Odyssey, Mystery Fun House, Ghost Town and The Golden Voyage to completion.

I'd say this is ready to go.

cspiegel:
Sorry for the long delay, but I'd say it's high time to get this in!

A couple considerations:

1. I think it'd probably be good to have this in a separate upstream repository, so it can be properly shared by both Spatterlight and Gargoyle (and anybody else interested).  I know that's making a bit more work for you: is it something you're willing to do? If so, we can then merge it in a single commit as with other interpreter updates.
2. Does this have 100% the same behavior as the original scottfree interpreter for games that it (the original) supports?  I'm just wondering if there's a reason to keep the old interpreter around (disabled by default) in case there are games it's needed for.  If not, then a wholesale replacement is fine.

angstsmurf:
> 1. I think it'd probably be good to have this in a separate upstream repository, so it can be properly shared by both Spatterlight and Gargoyle (and anybody else interested).

The current ScottFree code in Spatterlight is very different from this. The major part of it is support for the ZX Spectrum and Commodore 64 cassette versions released by Adventure International UK, with graphics and animations, none of which is included here. The point of this pull request was to start small with a subset that was easier to review, leaving out all the more hacky and controversial stuff, such as the 6502 emulator used to decompress the Commodore 64 versions, and then cleaning up and adding the rest later.

By the way, someone just ported the complete Spatterlight ScottFree interpreter, 6502 emulator, German grammar and all, to [C++ and the ScummVM framework](https://github.com/scummvm/scummvm/tree/master/engines/glk/scott). It might even make more sense to port that to Gargoyle instead, as some cruft was removed in the porting process, but that would require a bit of work.

None of this precludes making a separate repository out of this pull request, of course.

> 2. Does this have 100% the same behavior as the original scottfree interpreter for games that it (the original) supports? 

Well, yes and no. The parser works a bit differently, but if you enter the exact same commands the output should be identical.

cspiegel:
> The current ScottFree code in Spatterlight is very different from this. The major part of it is support for the ZX Spectrum and Commodore 64 cassette versions released by Adventure International UK, with graphics and animations, none of which is included here. The point of this pull request was to start small with a subset that was easier to review, leaving out all the more hacky and controversial stuff, such as the 6502 emulator used to decompress the Commodore 64 versions, and then cleaning up and adding the rest later.

Ok, I see.

> By the way, someone just ported the complete Spatterlight ScottFree interpreter, 6502 emulator, German grammar and all, to C++ and the ScummVM framework. It might even make more sense to port that to Gargoyle instead, as some cruft was removed in the porting process, but that would require a bit of work.

I've looked at ScummVM's Glk implementation before, and they really hack up the interpreters to fit into their model. To port it over we'd need a compatibility layer, at least providing the common routines they use (`Common::Array`, etc).  Maybe that's worth it, maybe not... and it looks like they at least hook into the Scumm VM save system, which would have to be adapted.

Offhand I think it'd be easier to get the Spatterlight code extracted and usable everywhere, but I'll try to find some time to see just how difficult it'd be to use the ScummVM code with a shim layer.  For now, I'm not against a separate "in between" version as represented by this pull request.

> None of this precludes making a separate repository out of this pull request, of course.

I do think this is the right approach: I'd prefer for Gargoyle to not be the "official" source for any interpreters.  At the moment the Scott Adams interpreter is from https://github.com/cspiegel/scottfree-glk, so it has an official-esque upstream, although it's just my straight port of the original ScottFree to Glk.  I'm slowly trying to decouple Gargoyle from the interpreters so that it's easier to see Gargoyle's modifications, and to upgrade interpreters when necessary.  A new upstream Scott Adams interpreter would make things a lot cleaner, and I'd rather keep my scottfree-glk repo just the original 1.14 ported to Glk, with no embellishments.

If you'd prefer I can create a repository under the garglk organization to hold this version of the interpreter, and features from Spatterlight can slowly be added to it, then upstreamed here.  If you're in favor of that I can do the work of shifting this branch over as the initial commit there.

>> Does this have 100% the same behavior as the original scottfree interpreter for games that it (the original) supports?

> Well, yes and no. The parser works a bit differently, but if you enter the exact same commands the output should be identical.

OK, that's reasonable enough to not need two interpreters.

angstsmurf:
I'm closing this now. It's too out of sync with the Spatterlight version. I made a fork here with all the three Scott Adams-related interpreters (ScottFree, TaylorMade and Plus) running in Gargoyle:

https://github.com/angstsmurf/garglk/tree/saga

Seems to work fine apart from locking up on certain animations. And I don't know how to get the AppImage workflow to build.

PR #631, opened 2022-03-18
Provide a new glkunix_fileref_get_filename API function

cspiegel:
This is identical to the existing garglk_fileref_get_name function, but
for compatibility with remglk, the new name is used (and preferred).

The original function is marked as deprecated, but will remain in the
library for the foreseeable future.

### DIFF
--- a/garglk/cheapglk/cgfref.cpp
+++ b/garglk/cheapglk/cgfref.cpp
@@ -542,3 +542,7 @@ void glkunix_set_base_file(char *filename)
 #endif
 }
 
+const char *glkunix_fileref_get_filename(fileref_t *fref)
+{
+    return fref->filename;
+}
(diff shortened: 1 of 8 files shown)


cspiegel:
@curiousdannii I removed the TADS code to handle remglk, because if `GLKUNIX_FILEREF_GET_FILENAME` is defined, it includes `remglk.h` which doesn't exist, of course, in Gargoyle.

If this needs to support both the old garglk and new glkunix names, I _think_ the proper approach is something like this:

```c
#include "glk.h"
#include "glkstart.h"

#if defined(GLK_MODULE_FILEREF_GET_NAME) && !defined(GLKUNIX_FILEREF_GET_FILENAME)
#define glkunix_fileref_get_filename garglk_fileref_get_name
#define GLKUNIX_FILEREF_GET_FILENAME
#endif
```
That is, if the old name exists but the new name doesn't, provide an alias to the old name and define the feature test macro for it.  Then the TADS code can, as it already does, just check `GLKUNIX_FILEREF_GET_FILENAME`.

PR #634, opened 2022-03-22
Flush FILE streams on write

cspiegel:
When upstream cheapglk was brought in, some fflush() calls got lost,
resulting in (among other things, probably) transcripts not getting
written until the buffer filled up, or the file was closed, whichever
came first.

fflush() is now called whenever a character or buffer is written. This
is less efficient than it ought to be, since glk_put_buffer_uni() is now
implemented in terms of gli_put_char_uni(). That means that when writing
a buffer, fflush() is called for each and every character. Accept this
for now, since it's never going to be a huge bottleneck, and only TADS
is actually making use of glk_put_buffer_uni(). But at some point I'd
like to find a solution that doesn't cause the code to deviate too much
from upstream cheapglk, to keep future updates as painless as possible.

### DIFF
--- a/garglk/cheapglk/cgstream.c
+++ b/garglk/cheapglk/cgstream.c
@@ -741,6 +741,7 @@ static void gli_put_char(stream_t *str, unsigned char ch)
                     putc(ch, str->file);
                 }
             }
+            fflush(str->file);
             break;
         case strtype_Resource:
             /* resource streams are never writable */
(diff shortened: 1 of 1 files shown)


curiousdannii:
Would it be more efficient to flush just when glk_select is called? That would still result in effectively instant flushing.

cspiegel:
> Would it be more efficient to flush just when glk_select is called? That would still result in effectively instant flushing.

Yes, and ultimately that'd be the approach I want to take. This first fix is essentially a reversion to the original Gargoyle behavior. To flush on select would require iterating over all streams, which is more invasive, and I'd need to think more about any possible side effects.

But I do think it's the right approach for a proper fix!

PR #639, opened 2022-03-29
Fix usages of glkunix_fileref_get_filename

cspiegel:
Unlike the earlier Gargoyle-specific function, this one is defined in
glkstart.h, a Unix Glk-specific header (which includes garglk). As such,
this header must be included to get the GLKUNIX_FILEREF_GET_FILENAME
macro.

One small problem: both Alan2 and TADS want to use this functionality,
but also include their own copy of glkstart.h. There's no reason for
them to have this, since the Glk implementation ought to be the one
providing it. As such, these versions have been removed from the
repository.

This also means that a test against GARGLK is needed before including
glkstart.h. For portability this should be more than just GARGLK, but in
this repository, it's fine. For use outside of Gargoyle, this should be
more specific, e.g. testing a macro for whether it's Unix Glk, Windows
Glk, etc. Right now, GARGLK is a proxy meaning "Unix Glk", used because
these interpreters don't care about anything but Gargoyle.


cspiegel:
@curiousdannii @cibersheep This has some material changes to interpreters that you might be interested in seeing.  It's noted above, but the point here is that the new GLKUNIX_FILEREF_GET_FILENAME test is in `glkstart.h` as it's part of the glkunix API, so `glkstart.h` must be included; and since it's for Unix Glk, it's not guaranteed to exist.  I assume it does if GARGLK is set, but that's not a general solution.

In addition, a few interpreters include their own `glkstart.h`, but of course it's not tailored to Gargoyle, so I've removed those which clash.

cibersheep:
> In addition, a few interpreters include their own `glkstart.h`, but of course it's not tailored to Gargoyle, so I've removed those which clash.

Thanks for the headsup

curiousdannii:
Changing the constants used in TADS is something I've already been planning to do: https://github.com/tads-intfic/tads-runner/issues/11

cspiegel:
Just as a note, I've approached this in Bocfel as follows:

One macro is set if Glk support is requested.
If Glk support is requested, another macro is set indicating which Glk implementation is being used (as of right now, that's just Unix and Windows).

So the code looks something like this:

```c
#ifdef GLK
#include <glk.h>
#if defined(GLKUNIX)
#include <glkstart.h>
#elif defined(GLKWINDOWS)
#include <WinGlk.h>
#else
#error Unsupported Glk platform
#endif
```
This is clearly ugly, but it's pretty easy to confine it.

PR #649, opened 2022-06-08
Fix loading of multimedia for JACL and Hugo

### LINKED ISSUE #648: Multimedia broken in JACL and Hugo
Due to a recent change to how files are named (and bringing Gargoyle in line with Glk's recommendations), JACL and Hugo lost the ability to load multimedia files.

The change was to unconditionally append `.glkdata` to files opened via Glk when used for data (text and save files have extensions added as well).  JACL and Hugo assumed that Glk could open arbitrary files, which Gargoyle _could_ do prior to this change.

Gargoyle will continue to follow the Glk spec and add `.glkdata`, so the interpreters need to be changed to either use `glkunix_stream_open_pathname` or directly use `stdio` routines, where appropriate.

cspiegel:
Fixes #648

### DIFF
--- a/terps/hugo/heglk/heglkunix.c
+++ b/terps/hugo/heglk/heglkunix.c
@@ -82,6 +82,10 @@ glkunix_startup_error (char *fmt, ...)
     }
   va_end (ap);
 }
+
+#ifdef GARGLK
+char *hugo_path_to_game;
+#endif
   
 int
 glkunix_startup_code (glkunix_startup_t *data)
@@ -109,6 +113,22 @@ glkunix_startup_code (glkunix_startup_t *data)
     s = strrchr(data->argv[1], '/');
     if (!s) s = strrchr(data->argv[1], '\\');
     garglk_set_story_name(s ? s + 1 : data->argv[1]);
+
+    hugo_path_to_game = calloc(1, strlen(data->argv[1]) + 1);
+    if (hugo_path_to_game == NULL)
+      {
+        glkunix_startup_error ("Error: out of memory");
+        return FALSE;
+      }
+
+    if (s != NULL)
+      {
+        memcpy(hugo_path_to_game, data->argv[1], s - data->argv[1]);
+      }
+    else
+      {
+        strcpy(hugo_path_to_game, data->argv[1]);
+      }
 #endif
 
   glkunix_set_base_file(data->argv[1]);
--- a/terps/hugo/heglk/hemedia.c
+++ b/terps/hugo/heglk/hemedia.c
@@ -73,6 +73,33 @@ static int loadres(HUGO_FILE infile, int reslen, int type)
 		return -1;
 
 	id = numres[type]++;
+#ifdef GARGLK
+	// Gargoyle now appends ".glkdata" to files opened with normal Glk file
+	// functions. That means that this would create files such as
+	// "PIC0.glkdata", but Gargoyle requires names like "PIC0". If Gargoyle
+	// used Glk routines to open individual media files, then Glk routines
+	// could be used to create them, but it doesn't; instead, it uses stdio
+	// routines, so do the same here.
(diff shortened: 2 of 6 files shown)


cspiegel:
> It would be good to get the Hugo changes upstream. And do they all need to be gated behind GARGLK checks? I don't see any GARGLK specific functions.

A couple things...

First, the technically-not-standard `glkunix_stream_open_pathname` function is used, meaning this is limited to Glk implementations which provide a glkunix startup, which, to be fair, is most of them.  This _could_ be rewritten to use standard I/O routines, but the way the interpreter is written means that's a bit invasive: macros with the same name as standard I/O functions are defined as wrappers around Glk functionality, presumably because the interpreter originally was written to use standard I/O routines only, and Glk support was retrofitted via these wrappers. Anyway, it makes a "proper" fix more difficult.  For now the easy fix is using this glkunix function, but I'm not sure how upstream would prefer that be handled, since it's not a standard Glk function.  Maybe it's enough to just say "we support Unix startups only".  I don't know.  The alternative, I suppose, would be requiring a macro to be set at build time which is just pushing the work onto the person doing the building, but then, that person knows which Glk implementation is being used, so it's not a huge burden (this is how I do it with Bocfel).

The next part, though, is tied to specific Glk implementations.  Multimedia support is very hacky: pictures/sounds are extracted from the Hugo resources and written directly to files with names like `PIC0` and `SND0`, which Gargoyle is able to load if there's not a Blorb file associated with the game.  xglk supports these PIC files, though I'm not sure any others do...  This is a "suggested" method by the Blorb specification, but I don't know how appreciated it'd be if the Hugo interpreter dumped these files next to every Hugo game regardless of whether the Glk implementation supported them.  In all honesty, though, I'm not sure any other Glk implementations can currently use Hugo multimedia resources, which means it may indeed make sense to hide these behind `GARGLK`, which would then also mean that the code using `glkunix_stream_open_pathname` would still want to be protected by `GARGLK` as well, since there's no use in even looking at the resource files if they can't even be used.

I admit I wasn't thinking about upstream when I hacked this solution together, and there is one thing that I'll definitely clean up for that, and just because it's the right thing to do: the new `hugo_path_to_game` variable isn't declared in any headers, but instead directly `extern`'d from source files. This is awful practice and I shouldn't have done it.  So even if it never gets upstreamed, I'll fix it anyway, because it's not cool.

I think I'll probably have to open an issue upstream to forge a path forward there.  For now I'll leave this pull request in stasis, but one way or another I'll ensure the fix gets in soonish, to try to maybe push out a 2022.1.2 release.

curiousdannii:
Ahh, looks like glkstart.h doesn't have its own header to rely on? Make, like Bocfel, the terps should have a GLK_UNIX define then. And if someone wants to build it against Windows Glk for example, then they can add their own parallel solution.

> pictures/sounds are extracted from the Hugo resources and written directly to files with names like PIC0 and SND0, which Gargoyle is able to load if there's not a Blorb file associated with the game. xglk supports these PIC files, though I'm not sure any others do... This is a "suggested" method by the Blorb specification, but I don't know how appreciated it'd be if the Hugo interpreter dumped these files next to every Hugo game regardless of whether the Glk implementation supported them. In all honesty, though, I'm not sure any other Glk implementations can currently use Hugo multimedia resources, which means it may indeed make sense to hide these behind GARGLK, which would then also mean that the code using glkunix_stream_open_pathname would still want to be protected by GARGLK as well, since there's no use in even looking at the resource files if they can't even be used.

Ah, yeah that's quite hacky. A better approach, I think, could be to construct an in-memory blorb. Though I haven't tried it.

cspiegel:
I've got an issue open upstream at hugoif/hugo-unix#4, and if that gets resolved, I'll update from there.  For now, though, ensure loading is fixed locally for future releases.

PR #664, opened 2022-07-18
Support color themes

cspiegel:


cspiegel:
This may well need some tweaking, including possibly changing the format of the themes, but the general idea is there.

PR #691, opened 2022-12-09
New and updated Scott Adams-related interpreters

angstsmurf:
This PR adds the [TaylorMade](https://www.ifwiki.org/Taylormade) and [Plus](https://intfiction.org/t/spatterlight-now-with-colour-cycling-animations/57417) interpreters to Gargoyle, and updates ScottFree to support graphics in many games, along with the other improvements from #627.

cf07f08 fixes the lock-up problem previously seen during some animations on macOS. I haven't testes this PR on other operating systems.

In Spatterlight, Unp64 (for decompressing Commodore 64 releases) and c64diskimage (for reading Commodore 64 disk images) are shared dynamic libraries. I don't know how to do that in a cross-platform way, so here they are compiled twice or thrice along with each individual interpreter. There is more duplicated code that could potentially be shared.

TaylorMade has been ported from from Alan Cox's original code to Glk and now supports graphics and _Questprobe 3_. It also fixes some original bugs that made games uncompletable.

_Rebel Planet_ emits some beeps that might be possible to reproduce witch the Gargoyle Z-Machine bleep functionality in a future update.

TaylorMade only supports the ZX Spectrum and Commodore 64 versions of the games, and can read files in z80, sna, tzx, tap, t64, and d64 formats, but all files must be renamed and given a `.tay` extension for Gargoyle to open them.

Plus supports all known S.A.G.A Plus games, in d64, dsk, atr, st, msa, and dat format, but all must be renamed and given a `.plus` extension.

Many Plus images have ugly edge artifacts that were not visible against the black background of the originals. I might look into cropping these in the future.

The updated ScottFree supports graphics in the UK cassette releases for Commodore 64 and ZX Spectrum, including the Mysterious Adventures series. It also supports graphics in many later S.A.G.A releases, except those that used vector graphics and _Return to Pirate's Island_. It can read the z80, sna, tzx, tap, t64, d64, dsk, woz, atr, dat, and fiad formats, but all must be given a `.saga` extension. For games distributed as two disk image files, both files must be renamed.

The unicode characters used to draw the borders and dice on the battle screen of _Seas of Blood_ show up as question marks in Gargoyle. I haven't checked whether this is just a font problem or they are outside some Glk spec.


curiousdannii:
Does it really do animations in pure Glk just through `glk_window_fill_rect`? I've thought a little bit before about adding a Glk extension for drawing a bitmap from memory, which would surely help with this sort of thing.

angstsmurf:
Yes. This is not the most efficient way to do it, but even if you could draw a bitmap from memory, you would still have to create the bitmap pixel by pixel in the first place, so the difference may not be that great. The big gain would be the ability cache the final bitmaps and reuse them without having to redraw them every time.

cspiegel:
Pay no attention to the closing/opening, it's just somebody who can't tell the difference between buttons.

angstsmurf:
> Are any of the newly-supported games available anywhere with easy access?

Most of them can be found on the Internet Archive. A simple search for [scott adams saga](https://archive.org/search.php?query=scott+adams+saga) will give a lot of hits, perhaps too many. The hard part is finding the ones that work.

cspiegel:
I have a small pull request at angstsmurf/garglk#3.  With that I think we can get this merged in.

I also verified that the box drawing characters in Seas of Blood do work with a font that has Unicode box-drawing support.

angstsmurf:
Some more links with games for testing:

[Games in TI-99/4A Scott Adams format at If-archive](https://www.ifarchive.org/indexes/if-archive/scott-adams/games/ti99/)
[C64 Mysterious Adventures at If-archive](https://www.ifarchive.org/if-archive/scott-adams/games/c64/mystadv.zip)
[Atari ST Questprobe 2 at Atarimania](http://www.atarimania.com/game-atari-st-questprobe-ii-spiderman_28598.html)
[Atari ST Questprobe 3 at Atarimania](http://www.atarimania.com/game-atari-st-questprobe-iii-fantastic-four-chapter-i_28597.html)
[Atari ST Sorcerer of Claymorgue Castle at Atarimania](http://www.atarimania.com/game-atari-st-saga-13-sorcerer-of-claymorgue-castle_31317.html)
[Apple 2 The Count at the Internet Archive](https://archive.org/download/wozaday_SAGA_5_The_Count)
[Apple 2 Voodoo Castle at the Internet Archive](https://archive.org/download/wozaday_SAGA_4_Voodoo_Castle)

cspiegel:
Things are looking good to me.  Thanks for this!

Any reason to hold up the merge, or do you consider it ready now?

cspiegel:
One thing I didn't think of: can you add the new interpreters to the README, so they show up on the landing page?

Another thing that may be more problematic: licenses.  First, there's no license in the "plus" interpreter, but since you wrote it (apart from the Atari drawing code, which is MIT, so is fine), I imagine that's easy to resolve.

But the unp64 code has this in the license:

> Permission is granted to anyone to use this software, alter it and re-distribute it freely for any non-commercial, non-profit purpose subject to the following restrictions:

I believe this is incompatible with the GPL, since it restricts commercial/for-profit use of the code, which the GPL allows.  Basically, it's putting restrictions on the code beyond what the GPL does, and the GPL doesn't allow that.  I'm not sure the right way to go here.  Unfortunately unp64 looks to be pretty big, meaning a clean rewrite isn't much of an option.

angstsmurf:
Right, I remember now that I was worried that would be a problem when they ported ScottFree to ScummVM.  Of course it is a problem for Spatterlight as well, but I just didn't consider it before. I guess the ScummVM people either didn't notice the license or thought that rewriting it in C++ was enough to circumvent it.

I've mailed the author of UNP64, [Ian Coog](http://iancoog.altervista.org), and asked whether he would consider releasing it under a GPL-compatible license. For now I'll just remove the parts that rely on it, which is the European cassette-based C64 releases (mostly available only as D64 disk images). There are a few of those that have unprotected versions as well, and those will work without it.

angstsmurf:
I removed UNP64, and it seems to work.

Ian Coog replied that we can do whatever we want with the code as long as we don't remove the original names from the sources, but that it is based on another tool, [Exomizer](https://bitbucket.org/magli143/exomizer/src/master/)  by Magnus Lind, so I've written to him as well.

cspiegel:
Great, thank you!  I'm merging this in.  We can always restore UNP64 later on if we're given the green light to relicense under something compatible with the GPL.

PR #714, opened 2023-01-05
ScottFree, Plus, TaylorMade: Minor fixes from upstream

angstsmurf:
These are a number of small fixes from Spatterlight. Let me know if you prefer separate PRs, if commits should be squashed, or if there is anything else you would like me to change.


cspiegel:
All at once is just fine.  Since you're the maintainer of these, I can be certain that they're being brought in properly, however you choose to do it.  Feel free to create pull requests whenever you feel it's appropriate, as you know when changes are necessary.

At some point, when the dust settles a bit on these new interpreters, it may be best to treat Spatterlight as the official source of all Scott-related interpreters, and upstream from it whenever there's a new release, same as any other interpreter.

PR #715, opened 2023-01-06
PLUS, SCOTT: Fix use of snprintf()

angstsmurf:
Hopefully this is better.

### DIFF
--- a/terps/plus/loaddatabase.c
+++ b/terps/plus/loaddatabase.c
@@ -777,8 +777,9 @@ static int SetGame(const char *id_string, size_t length) {
 
 int FindAndAddImageFile(char *shortname, struct imgrec *rec) {
     int result = 0;
-    char filename[DirPathLength + 9];
-    int n = snprintf(filename, sizeof filename, "%s%s.PAK", DirPath, shortname);
+    size_t pathlen = DirPathLength + 9;
+    char *filename = MemAlloc(pathlen);
+    int n = snprintf(filename, pathlen, "%s%s.PAK", DirPath, shortname);
     if (n > 0) {
         FILE *infile=fopen(filename,"rb");
         if (infile) {
@@ -799,6 +800,7 @@ int FindAndAddImageFile(char *shortname, struct imgrec *rec) {
             debug_print("Could not find or read image file %s\n", filename);
         }
     }
+    free(filename);
     return result;
 }
 
--- a/terps/plus/loaddatabase.c
+++ b/terps/plus/loaddatabase.c
@@ -777,8 +777,8 @@ static int SetGame(const char *id_string, size_t length) {
 
 int FindAndAddImageFile(char *shortname, struct imgrec *rec) {
     int result = 0;
-    char filename[2048];
-    int n = snprintf(filename, DirPathLength + 9, "%s%s.PAK", DirPath, shortname);
+    char filename[DirPathLength + 9];
+    int n = snprintf(filename, sizeof filename, "%s%s.PAK", DirPath, shortname);
     if (n > 0) {
         FILE *infile=fopen(filename,"rb");
         if (infile) {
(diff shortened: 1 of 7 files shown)


angstsmurf:
Missed some `snprintf()` in TaylorMade.

terps/plus/loaddatabase.c
```diff
@@ -777,8 +777,8 @@ static int SetGame(const char *id_string, size_t length) {
 
 int FindAndAddImageFile(char *shortname, struct imgrec *rec) {
     int result = 0;
-    char filename[2048];
-    int n = snprintf(filename, DirPathLength + 9, "%s%s.PAK", DirPath, shortname);
+    char filename[DirPathLength + 9];
```
cspiegel, line 780: Creating an array at runtime (i.e. a variable-length array, or VLA) like this may have consequences; points to consider:

* It's not legal in C89/C90 (but snprintf() is not there either, so that's no big deal)
* It's guaranteed to be supported in C99
* It's _not_ guaranteed to be supported in C11 (it was basically made optional)

It's also the one time that `sizeof` _is_ computed at runtime, but that's not a huge deal.

I have zero problems with VLAs as a concept, and used them in Bocfel, removing them only when I moved to C++, which doesn't support them.  However, Gargoyle _does_ force all interpreters to be built in C11 mode.  It would be pretty easy to allow the standard to be selected and then just use C99 for plus.  The alternative would be a fixed-size array as it used to be.

There is, unfortunately, no bulletproof/ideal solution here (typical of C, unfortunately).  I'm OK with adding support to Gargoyle for interpreters to select which C standard they adhere to, which is already possible for C++ interpreters.

terps/scott/saga/pcdraw.c
```diff
@@ -28,9 +28,8 @@ extern char *DirPath;
 uint8_t *FindImageFile(const char *shortname, size_t *datasize) {
     *datasize = 0;
     uint8_t *data = NULL;
-    char filename[2048];
-    size_t namelength = strlen(DirPath) + strlen(shortname) + 5;
-    int n = snprintf(filename, namelength, "%s%s.PAK", DirPath, shortname);
+    char filename[strlen(DirPath) + strlen(shortname) + 5];
```
cspiegel, line 31: As above re VLAs.

terps/plus/loaddatabase.c
cspiegel: Let me here play devil's advocate against VLAs, though, so you have the full picture.  The argument against them typically goes like this:

Taking up too much automatic storage (i.e. "the stack" in every practical system) is undefined behavior.  If your VLA can take on any size, then you run the risk of crashing, since your array will overflow the stack.  Therefore you need to ensure your VLA is _not_ unbounded in size, but instead can only grow to a reasonable maximum.  Since your VLA can only grow to a reasonable maximum, just create a fixed-size array of that size.

angstsmurf: So isn't it a viable workaround to just use `malloc()` instead? You'd think that would be safer, as we can check if it fails and exit, rather than risking writing outside the end of an array.

curiousdannii: Want to rewrite in Rust? 😛

cspiegel: > Want to rewrite in Rust? stuck_out_tongue

The main reason Bocfel is C++ and not Rust is for compatibility with all platforms that Gargoyle supports...

Ok, that's not _entirely_ true but it's certainly not entirely false, either!

angstsmurf: I'm have no idea how long a file path string passed as an argument to a program can be, but that would be the upper limit of the size in this case. If it can be huge, then `malloc()` is the better alternative here I suppose.

cspiegel: > I'm have no idea how long a file path string passed as an argument to a program can be, but that would be the upper limit of the size in this case. If it can be huge, then `malloc()` is the better alternative here I suppose.

Implementations in the past have tried to set values for that (MAX_PATH or PATH_MAX, for example), but they're not really ideal, since some systems allow arbitrary (or close-to) values for the math value of a pathname, and those macros are then just lies.

`malloc()` is definitely the way to ensure that you have a large enough buffer since you can't know how large `argv[n]` can get.  Some programs will also just set a value like 4096 and damn the user if they have such a ridiculous directory hierarchy.  Either's a fine approach, as far as I'm concerned.

PR #723, opened 2023-01-09
SCOTT and TAYLOR: Reinstate Unp64

angstsmurf:
I finally got a reply from Magnus Lind, author of the Exomizer tool, which Unp64 borrows a lot of code from. He agrees to let us distribute his code under the zlib license, but it turns out neither of us is sure how to best do this formally. He supplied this statement:

>The copyright holder of the Exomizer code, Magnus Lind, hereby dual licenses the code from Exomizer used in unp64. It is now also available under the Zlib-license which makes it compatible with the GPL. (https://en.wikipedia.org/wiki/Zlib_License).

I've tried to update the copyright text in the source and updated our version of Unp64 to work with the latest TaylorMade and ScottFree changes. There is probably still too much code left in there that just prints information to the console and can't be switched off.

### DIFF
--- a/terps/CMakeLists.txt
+++ b/terps/CMakeLists.txt
@@ -64,6 +64,10 @@ if(WITH_PLUS OR WITH_SCOTT OR WITH_TAYLOR)
     add_subdirectory(c64diskimage)
 endif()
 
+if(WITH_SCOTT OR WITH_TAYLOR)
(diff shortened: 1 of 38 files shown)


cspiegel:
Ok, thanks for taking care of this! That copyright/license statement is good enough for me.

Issue #679, opened 2022-08-26
Glitchy graphical flicker when resizing window


salty-horse:
Using latest master on Linux.

When resizing the Gargoyle window, the contents of the window get smeared for a moment before redrawing correctly.

It started happening since commit 1ac6aeea54ffea08589263c95912515d7febf6d9 - Don't redraw directly inside of paint events.
![glitch](https://user-images.githubusercontent.com/28355/186929821-a300cc34-fe08-4928-8758-2b30df8c7ea6.png)

cspiegel:
This seems to be because the canvas is being redrawn before it's resized, i.e. the window gets resized, draws the old canvas, the canvas gets resized, then is redrawn. If you resize vertically only, the artifacts don't occur, because the horizontal lines remain the same size.

There's a lot of moving parts to the redrawing, and I've not really done much with them, so I'll have to figure out how they go together. But this ought to be fairly easy to fix.

angstsmurf:
Just another data point: On macOS, the sheared picture isn't just momentary. It appears when the user resizes the game window, and stays until it is resized again.

<img width="697" alt="Skärmavbild 2022-10-11 kl  02 08 44" src="https://user-images.githubusercontent.com/5066755/194970608-6fc86d9f-3549-41a1-a011-116c6e1c3aec.png">

cspiegel:
Hm... ok, that's an issue.  The Mac display code is different than the Qt display code, but I modeled this part of the Qt code on the Mac code so I suppose it makes sense they'd exhibit similar behaviors.

It's really a pain dealing with the resizing signals and threading all that information through at the right time, but I'll try to get a handle on it.

Nephar:
I can reproduce this issue without window resizing(the terp launches looking like below):

<img width="751" alt="Screen Shot 2023-01-14 at 10 26 38" src="https://user-images.githubusercontent.com/122389007/212477087-2dd788c1-d995-44e3-80f4-ce461ceaad6f.png">

To reproduce it, only 2 changes to the default garglk.ini are needed.
1. increase cols to 61
2. increase monosize to 18

Both changes are for global settings. I believe with that column width, monosize between 17.6-19.2 causes this screen. This bug might only present itself under Mojave, since that is what I am using.

angstsmurf:
I can confirm that the same thing happens on macOS 13.1 Ventura if I make those changes to the default settings.

cspiegel:
[...]

PR #731, opened 2023-01-14
Prevent distorted image on Mac by forcing window width to even

angstsmurf:
~~It feels a little silly to make a pull request for such a tiny change, but this is meant as a starting point for finding a fix for #679. This one-liner seems to prevent the shearing from happening at startup if `cols` is set to an odd value, but it will still occur if the window is resized manually to an odd width. I tried adding a similar line to the `winresize()`, but it didn't help.~~

Edit: Adding a delegate method that makes sure that all window size changes end up with an even width fixes it completely for me. Of course, there is probably some deeper cause of this that we should fix instead.

### DIFF
--- a/garglk/launchmac.mm
+++ b/garglk/launchmac.mm
@@ -613,6 +613,12 @@ static BOOL isTextbufferEvent(NSEvent *evt)
     return self;
 }
 
+- (NSSize)windowWillResize:(NSWindow *)sender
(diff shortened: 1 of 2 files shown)


angstsmurf:
Superseded by [this](https://github.com/garglk/garglk/issues/679#issuecomment-1382904269), I think.

PR #735, opened 2023-01-24
Update jacl

cspiegel:
This is from the master branch. While I'd prefer to import a release, JACL has no updated releases, but the last release (years ago) has some code which causes Clang to issue gobs of warnings. Stuart Allen, JACL's creator, has fairly recently resurrected the JACL code on Github, and has incorporated changes which address these warnings. They're abundant enough (if generally harmless) that I think it's worth pulling from master to fix them.


angstsmurf:
Perhaps I should address this to Stuart Allen instead, but I get weird errors in _The Unholy Grail_ with this version of JACL, which breaks my walkthrough script. What am I missing? 

For example, at the very start:
```
>W
(The pool)
>N
(The control centre)
>TURN OFF LIGHT
ERROR: Attempt to execute unknown function “+turn_off”.
```

With the old version of the interpreter, this works and gives 10 points.
```
>TURN OFF LIGHT
You flick the light switch into the off position causing the main fluorescent light to go out.

Lisa is here, standing close by.

> examine screen
Two perpendicular fine white lines form a large cross hair on the screen. Markings along both the X and Y axes indicate longitude and latitude to within point zero one of an arc second.

With the current low level of lighting, you can just make out the rapidly fading afterimage of an unusually large blob at precisely -171deg14'02.30" longitude and -23deg34'86.34" latitude.

Watching you intently, Lisa points to the screen. "It's a whale, we get a lot of them out here."

[YOUR SCORE JUST WENT UP BY 10 POINTS]
```

cspiegel:
Thank you for running this through a test!  I started up a few JACL games and played a few turns, and things were fine, but clearly there's a problem here.  Stuart would definitely be the one to talk to: this is a "mindless" port, and I don't have any JACL knowledge. Needless to say, I won't be merging this in as-is.

angstsmurf:
Right, I posted an [issue report](https://github.com/DangarStu/JACL/issues/17) at his repo.

cspiegel:
I've worked around this by simply disabling these warnings for JACL; I'll come back to upstreaming JACL when it's got a fix for this bug.

PR #740, opened 2023-02-12
Update Plus, ScottFree, and TaylorMade again

angstsmurf:
There is a lot of changes in this, so let me know if you want me to split it up.

I apparently broke the UK cassette-based version of *Questprobe featuring the Hulk* in one of the earlier updates. Now it works again.

Otherwise, the biggest change is that I've switched to a C++ version of Unp64 based on the one in ScummVM. It is not very object oriented, but the source is a little cleaned up.

I've added support for many TaylorMade game variants found on the Internet Archive. This required changes to Unp64 and C64diskimage. ~~[One version](https://archive.org/download/d64_Temple_of_Terror_1987_U.S._Gold_h_FLT) of *Temple of Terror* has an annoying bug where it randomly won't start, but not if it is built without optimisations. The problem seems to be somewhere in Unp64. It is entirely possible that this can be fixed by going back to the plain C version.~~ EDIT: Fixed it, just needed to null terminate some strings.

Reading Apple II images in WOZ format should be faster.


terps/unp64/CMakeLists.txt
```diff
@@ -1,13 +1,15 @@
...
-    scanners/_Scanners.c scanners/1001card.c scanners/AbuzeCrunch.c
-    scanners/ActionPacker.c scanners/ByteBoiler.c scanners/Caution.c
-    scanners/CCS.c scanners/Cruel.c scanners/ECA.c scanners/Exomizer.c
-    scanners/Expert.c scanners/FinalSuperComp.c scanners/Intros.c
-    scanners/MasterCompressor.c scanners/Megabyte.c scanners/MrCross.c
-    scanners/MrZ.c scanners/PuCrunch.c scanners/Section8.c
-    scanners/TBCMultiComp.c scanners/TCScrunch.c scanners/XTC.c)
-
-c_standard(unp64 11)
+set (CMAKE_CXX_STANDARD 17)
```
cspiegel, line 1: For setting the C++ standard, please use Gargoyle's `cxx_standard` function:

```
cxx_standard(unp64 17)
```
This takes a target rather than setting a global value. It also _requires_ the standard, and turns off extensions, but that's not as important.

terps/unp64/CMakeLists.txt
```diff
@@ -1,13 +1,15 @@
...
+
+add_library(unp64 STATIC 6502_emu.cpp exo_util.cpp globals.cpp unp64.cpp
+    scanners/scanners.cpp scanners/abuze_crunch.cpp scanners/action_packer.cpp
+    scanners/byte_boiler.cpp scanners/caution.cpp scanners/ccs.cpp scanners/cruel.cpp
```
cspiegel, line 5: As long as you're in the file, this contains a trailing newline that can be removed.

terps/unp64/CMakeLists.txt
```diff
@@ -1,13 +1,15 @@
...
+    scanners/byte_boiler.cpp scanners/caution.cpp scanners/ccs.cpp scanners/cruel.cpp 
+    scanners/eca.cpp scanners/exomizer.cpp scanners/expert.cpp scanners/final_super_comp.cpp
+    scanners/intros.cpp scanners/master_compressor.cpp scanners/mc_cracken_comp.cpp
+    scanners/megabyte.cpp scanners/mr_cross.cpp scanners/mr_z.cpp scanners/pu_crunch.cpp
+    scanners/section8.cpp scanners/tbc_multicomp.cpp scanners/tcs_crunch.cpp
+    scanners/time_cruncher.cpp scanners/xtc.cpp)
+
+target_include_directories(unp64 INTERFACE ${CMAKE_CURRENT_SOURCE_DIR} PRIVATE PARENT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+
 warnings(unp64)
```
cspiegel, line 15: And again, since you'll already be in the file, an extraneous newline which can be removed.

cspiegel:
During the build I get these warnings (gcc):

```
/home/chris/github/garglk/terps/unp64/scanners/mr_cross.cpp: In function ‘void Unp64::scnMrCross(UnpStr*)’:
/home/chris/github/garglk/terps/unp64/scanners/mr_cross.cpp:66:68: warning: comparison of integer expressions of different signedness: ‘unsigned int’ and ‘int’ [-Wsign-compare]
   66 |                                 (*(unsigned int *)(mem + p + 0x36) == 0x60D8d000 + (p >> 8)) /*((p+0x3b)>>8)*/
      |                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
/home/chris/github/garglk/terps/unp64/scanners/final_super_comp.cpp: In function ‘void Unp64::scnFinalSuperComp(UnpStr*)’:
/home/chris/github/garglk/terps/unp64/scanners/final_super_comp.cpp:162:65: warning: comparison of integer expressions of different signedness: ‘unsigned int’ and ‘int’ [-Wsign-compare]
  162 |                                 if ((*(unsigned int *)(mem + p) == ((((mem[0x2aa] + 0x15) & 0xff) << 24) | 0x00B99EA0)) &&
      |                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/chris/github/garglk/terps/unp64/unp64.cpp: In function ‘int Unp64::unp64cpp(uint8_t*, size_t, uint8_t*, size_t*, const char*)’:
/home/chris/github/garglk/terps/unp64/unp64.cpp:97:20: warning: ‘char* strncpy(char*, const char*, size_t)’ output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
   97 |             strncpy(string, switches, string_length);
      |             ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/chris/github/garglk/terps/unp64/unp64.cpp:94:50: note: length computed here
   94 |         Unp64::size_t string_length = std::strlen(switches);
      |                                       ~~~~~~~~~~~^~~~~~~~~~
/home/chris/github/garglk/terps/unp64/unp64.cpp:102:24: warning: ‘char* strncpy(char*, const char*, size_t)’ output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
  102 |                 strncpy(settings[numSettings], setting, string_length);
      |                 ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/chris/github/garglk/terps/unp64/unp64.cpp:101:44: note: length computed here
  101 |                 string_length = std::strlen(setting);
      |                                 ~~~~~~~~~~~^~~~~~~~~
/home/chris/github/garglk/terps/unp64/unp64.cpp:591:24: warning: ‘char* strncpy(char*, const char*, size_t)’ output may be truncated copying 248 bytes from a string of length 259 [-Wstringop-truncation]
  591 |                 strncpy(name, forcedname, 248);
      |                 ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
```
I will add individual comments to the code regarding these, since I think it'll be easier than dumping the warnings all over the place.

terps/unp64/scanners/mr_cross.cpp
```diff
@@ -0,0 +1,100 @@
...
+         different address than $0801
+         */
+		p = unp->_info->_run;
+		if (p == -1)
+			p = unp->_info->_start;
+		for (q = p + 0x80; p < q; p++) {
+			if ((*(unsigned int *)(mem + p + 0x00) == 0x8538A978) &&
+				(*(unsigned int *)(mem + p + 0x04) == 0x9AF9A201) &&
+				(*(unsigned int *)(mem + p + 0x0b) == 0xCA01069D) &&
+				(*(unsigned int *)(mem + p + 0x36) == 0x60D8d000 + (p >> 8)) /*((p+0x3b)>>8)*/
```
cspiegel, line 66: This gives a "signed/unsigned" comparison in g++, and it's beacuse of the addition of `(p >> 8)`.  Remove that, and it's fine, which is strange, because it's still comparing an unsigned int to an int in either case.  But anyway, this can be worked around by giving the constant a `U` suffix to make it unsigned, i.e. `0x60D8d000U + (p >> 8)`.

terps/unp64/scanners/final_super_comp.cpp
```diff
@@ -0,0 +1,284 @@
...
+		}
+	}
+	/* FinalSuperCompressor/flexible hacked? 2nd layer */
+	if (unp->_depAdr == 0) {
+		if (unp->_info->_start == 0x2a7) {
+			if (mem[0x2a7] == 0x85 &&
+				mem[0x2a8] == 0x01 &&
+				mem[0x2a9] == 0x4c) {
+				p = READ_LE_UINT16(&mem[0x2aa]);
+				if ((*(unsigned int *)(mem + p) == ((((mem[0x2aa] + 0x15) & 0xff) << 24) | 0x00B99EA0)) &&
```
cspiegel, line 162: Another signed/unsigned comparison warning in g++.  This can be addressed by making this constant unsigned, i.e. `0x00B99EA0U`.

terps/unp64/unp64.cpp
```diff
@@ -0,0 +1,745 @@
...
+
+int unp64cpp(uint8_t *compressed, size_t length, uint8_t *destinationBuffer, size_t *finalLength, const char *switches) {
+
+    char settings[4][64];
+    int numSettings = 0;
+
+    if (switches != NULL) {
+        Unp64::size_t string_length = std::strlen(switches);
+        char string[100];
+        if (string_length > 0 && string_length <= 100) {
```
cspiegel, line 96: This will pass if `string_length` is 100, and then effectively do this on line 98: `string[100] = '\0'`; but `string` is only 100 bytes long, so that will overflow it.

terps/unp64/unp64.cpp
```diff
@@ -0,0 +1,745 @@
...
+            strncpy(string, switches, string_length);
+            string[string_length] = '\0';
+            char *setting = strtok(string, " ");
+            while (setting != NULL && numSettings < 4) {
+                string_length = std::strlen(setting);
+                strncpy(settings[numSettings], setting, string_length);
+                settings[numSettings][string_length] = '\0';
```
cspiegel, line 103: Similar to above:

```c
strncpy(settings[numSettings], setting, sizeof settings[numSettings]);
settings[numSettings][(sizeof settings[numSettings]) - 1] = '\0';
```

terps/unp64/unp64.cpp
```diff
@@ -0,0 +1,745 @@
...
+			else {
+				if (p >= 0xffff)
+					_G(_unp)._lfMemF = 0 | _G(_unp)._ecaFlg;
+				break;
+			}
+		}
+	}
+
+	if (*forcedname) {
+		strncpy(name, forcedname, 248);
```
cspiegel, line 591: g++ also warns here.  Is there a particular reason in limiting to 248 bytes here?  In the `else` below it looks like the 248 limit is to ensure that `".%04x%s"` will have enough space to be appended, but that's not necessary here.  I'd think even `strcpy` would work here, since `name` and `forcedname` are the same size, though to be safer:

```c
snprintf(name, sizeof name, "%s", forcedname);
```
would be better.

terps/unp64/unp64.cpp
```diff
@@ -0,0 +1,745 @@
...
+
+    char settings[4][64];
+    int numSettings = 0;
+
+    if (switches != NULL) {
+        Unp64::size_t string_length = std::strlen(switches);
+        char string[100];
+        if (string_length > 0 && string_length <= 100) {
+            strncpy(string, switches, string_length);
+            string[string_length] = '\0';
```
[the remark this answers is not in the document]
cspiegel: Though let me also advocate `snprintf` here:

```c
snprintf(string, sizeof string, "%s", switches);
```
Easier to get right, since you don't need to bother with null termination manually.

terps/unp64/unp64.cpp
cspiegel: And `snprintf` here, too.

angstsmurf:
I think I've taken care of all of them now. Sorry about the force pushes.

angstsmurf:
Speaking of warnings, running this with UndefinedBehaviorSanitizer on gives lots and lots of warnings like
``
garglk/terps/unp64/scanners/abuze_crunch.cpp:82:9: runtime error: load of misaligned address 0x00016b4122bb for type 'unsigned int', which requires 4 byte alignment
``
and
``
garglk/terps/unp64/scanners/eca.cpp:104:12: runtime error: load of misaligned address 0x00016b412397 for type 'unsigned short', which requires 2 byte alignment
``
in all the scanners files. I have no idea of how to fix those other than to memcopy to new malloc'ed variables rather than assigning, and that would add a lot of lines and complexity.

angstsmurf:
The warnings seem to be gone. Fixed a number of out-of-bounds reads as well. Tested with a lot of games, and they all at least start.

cspiegel:
@angstsmurf Good to merge this in now, or is there something else you'll need to add here?

angstsmurf:
I think it is good. Nothing to add for now.

PR #769, opened 2023-03-18
PLUS, SCOTT, TAYLOR: Formatting

angstsmurf:
I've run clang-format on these files. I tried to match the Gargoyle style, but probably not hard enough. Let me know what needs to change.

Apart from looking nice, this will make it easier to merge changes from upstream.

### DIFF
--- a/terps/plus/animations.c
+++ b/terps/plus/animations.c
@@ -7,10 +7,12 @@
 
 #include <stdlib.h>
 #include <string.h>
+
+#include "common.h"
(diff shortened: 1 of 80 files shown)


cspiegel:
I'm sure there are still some niggling issues from when I recently standardized the Gargoyle source, so no worries if some stuff isn't there yet.  I've spent time, on and off, trying to create a clang-format style that works well, but it's soooo complex that I lose steam.  Eventually it may come, though!

PR #774, opened 2023-04-16
Update Fedora build instructions

Gragt:
Instructions to build Gargoyle on Fedora are slightly outdated. This commit includes a few missing dependencies and, more importantly, replaces `libjpeg-turbo-devel` with `turbojpeg-devel`. Both packages apparently provide the  same libjpeg-turbo library but the former identifies as libjpeg and the latter properly as libjpeg-turbo.

Because of Gargoyle’s recent changes to libjpeg requirements, it is currently impossible to build with `libjpeg-turbo-devel` despite it working in the past, but `turbojpeg-devel` allows it to build without problems.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -146,11 +146,11 @@ installer.
 
 # System-specific instructions
 
-## Building on Fedora GNU/Linux (29,30): (notes by Andreas Davour, December 2019)
+## Building on Fedora GNU/Linux (37, 38): (notes by Andreas Davour, December 2019; updated April 2023)
 
 To build cleanly with CMake, you first need to install the pre-requisite packages:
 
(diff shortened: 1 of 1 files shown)


cspiegel:
Thanks for the update!  One question: what's the libzip dependency for?  It's not directly used by Gargoyle; is there some transitive dependency situation here?

And while we're here, libpng is also required, which would be libpng-devel (I added most dependencies to the Fedora runner when I added the features, but just totally forgot that there's also a list in the build instructions!)

Gragt:
You are right: libzip-devel has no business there: it should have been zlib-devel—silly confusion on my part.

I did add libpng-devel from the beginning. I checked again, just to be sure.

I force pushed the changes; everything is hopefully correct now!

cspiegel:
Looks good, thanks!

PR #788, opened 2023-10-29
Support building with Clang on Windows

awlck:
Provide support for building Gargoyle using Clang on Windows.

Minor changes to the CMake build files. Minor changes to several interpreters in order to provide alternative implementations of POSIX functions that aren't available on Windows.

This PR does not yet contain a build script; that can be added later once we have interpreters that require using this configuration.

(For context, this is a prerequisite for the future inclusion of the FrankenDrift ADRIFT 5 interpreter via .NET AOT, which does not support cross-compilation. See: https://intfiction.org/t/frankendrift-play-adrift-games-on-mac-and-linux/51528/54)


garglk/cheapglk/cgdate.cpp
```diff
@@ -176,32 +179,47 @@ static glsi32 gli_simplify_time(time_t timestamp, glui32 factor)
 
 void glk_current_time(glktimeval_t *time)
 {
+#ifndef WIN32
```
cspiegel, line 182: My understanding is that `_WIN32` is defined by the compiler, and `WIN32` by headers, so that `_WIN32` is a bit less brittle; unless there's a strong reason to use `WIN32`, I'd prefer a blanket usage of `_WIN32` for consistency and robustness.

garglk/CMakeLists.txt
```diff
@@ -35,6 +35,8 @@ vtk_encode_string(
     NAME garglkini
     HEADER_OUTPUT GARGLKINI_H
     SOURCE_OUTPUT GARGLKINI_CXX
+    EXPORT_SYMBOL "GARGLK_API"
+    EXPORT_HEADER "garglk_importexport.h"
```
cspiegel, line 39: This doesn't work when building on MinGW:

```
In file included from /home/chris/github/garglk/garglk/config.cpp:58:
/home/chris/github/garglk/build-mingw/garglk/garglkini.h:4:10: fatal error: 'garglk_importexport.h' file not found

File: garglk/cheapglk/cgdate.cpp [hunk 3]
```diff
@@ -176,32 +179,47 @@ static glsi32 gli_simplify_time(time_t timestamp, glui32 factor)
...
     if (gettimeofday(&tv, NULL)) {
         gli_timestamp_to_time(0, 0, time);
         gli_strict_warning("current_time: gettimeofday() failed.");
         return;
     }
-
     gli_timestamp_to_time(tv.tv_sec, tv.tv_usec, time);
+#else
+    struct timespec ts;
+    if (!timespec_get(&ts, TIME_UTC)) {
```
cspiegel, line 192: `TIME_UTC` isn't defined on MinGW.  I'm guessing the proper approach is to check for MSVC (or for _not_ MinGW), and only do this for that platform.  MinGW should continue to use the POSIX interface.

garglk/cheapglk/cgdate.cpp
```diff
@@ -32,10 +32,13 @@
 #include <string.h>
 #include <stdlib.h>
 #include <time.h>
-#include <sys/time.h>
 #include "glk.h"
 #include "garglk.h"
 
+#ifndef _WIN32
+#include <sys/time.h>
+#endif
```
cspiegel, line 40: As below, I think this should be included for MinGW, and only removed with MSVC.

garglk/cheapglk/cgdate.cpp
```diff
@@ -176,32 +179,47 @@ static glsi32 gli_simplify_time(time_t timestamp, glui32 factor)
...
+    struct timeval tv;
     if (gettimeofday(&tv, NULL)) {
         gli_strict_warning("current_simple_time: gettimeofday() failed.");
         return 0;
     }
-
     return gli_simplify_time(tv.tv_sec, factor);
+#else
+    struct timespec ts;
+    if (!timespec_get(&ts, TIME_UTC)) {
```
cspiegel, line 217: As above.

terps/tads/glk/osfrobtads.h
```diff
@@ -34,7 +34,13 @@ extern "C++" {
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
+#define NOMINMAX
+#include <Windows.h>
```
cspiegel, line 40: This is another case of not building on MinGW, although it seems to be a case issue there. MinGW calls this `windows.h`, not `Windows.h`.  Since Windows is case insensitive, I guess it's safe to rename this as lowercase.

terps/tads/glk/osportable.cc
```diff
@@ -48,11 +48,14 @@
 #include "osextra.h"
 #include "charmap.h"
 #include <time.h>
+#ifndef _WIN32
```
cspiegel, line 51: This file comes from upstream frobtads, which is maintained.  Ideally this code would be upstreamed so that we don't have to patch it here in future releases. It's at least worth checking with Nikos to see if he's interested in adding this Windows support.  If not, no problem, we can continue to keep it here.  But as much as possible I want to avoid deviating from upstream interpreters.

garglk/cheapglk/cgfref.cpp
```diff
@@ -38,8 +38,17 @@
 #include <sys/stat.h> /* for stat() */
 
 #ifdef _WIN32
+#define _CRT_INTERNAL_NONSTDC_NAMES 1
+#define R_OK 02
```
cspiegel, line 42: This is a redefinition on MinGW, as it already provides `R_OK`, so this ought to at least check whether it's defined first.

cspiegel: Also, not sure of the effect it'd have, but on MinGW (at least the version I have installed), `R_OK` has the value 4, not 2, so there's a mismatch.

terps/CMakeLists.txt
```diff
@@ -121,7 +121,7 @@ endif()
 # Added glk file prompts for gargoyle using garglk_fileref_get_name.
 #
 if(WITH_ALAN2)
-    set(ALAN2_MACROS GLK)
+    set(ALAN2_MACROS GLK _PROTOTYPES_)
```
cspiegel, line 124: `_PROTOTYPES_` is already defined in `terps/alan2/sysdep.h`:

```
#ifdef __STDC__
#define _PROTOTYPES_
#include <stdlib.h>
#include <string.h>
#endif
```
This causes a redefinition warning.

`__STDC__` is required to be defined by C compilers; is MSVC, perhaps, not defining it?  If so, I'd think the best thing to do is add something in `sysdep.h` to check for MSVC in addition to `__STDC__`.

I've got a local experimental updated Alan2 interpreter which does away with all this pre-ANSI stuff, so hopefully in the future this won't even be a concern.

terps/tads/glk/osportable.cc
```diff
@@ -48,11 +48,14 @@
...
 #include <dirent.h>
+#endif
 #include <limits.h>
 
 #if defined(_WIN32)
 #include <windows.h>
+#define PATH_MAX MAX_PATH
```
cspiegel, line 58: MinGW already defines this, so it should only be done on MSVC.

terps/tads/glk/osportable.cc
cspiegel: Correction: this is from https://github.com/tads-intfic/tads-runner, not FrobTADS, though they share lots of code.

So it's worth seeing if _they_ are interested in these changes.

cspiegel:
One thing I'm noticing is that a lot of this new Win32 code will be used on MinGW as well.  That's not inherently a bad thing, just so long as the new Windows code works as well or better than the POSIX code.

The question kind of becomes, should MinGW and MSVC basically share most code, or should MinGW share as much with POSIX as possible?  I don't have an answer to that.

terps/tads/glk/osportable.cc
curiousdannii: Of course!

garglk/wingrid.cpp
```diff
@@ -77,8 +77,13 @@ void win_textgrid_rearrange(window_t *win, rect_t *box)
 
     for (k = 0; k < dwin->height; k++) {
         touch(dwin, k);
-        std::fill(&dwin->lines[k].chars[dwin->width], dwin->lines[k].chars.end(), ' ');
-        for (auto *attr = dwin->lines[k].attrs.begin() + dwin->width; attr != dwin->lines[k].attrs.end(); ++attr) {
+        // std::fill(&dwin->lines[k].chars[dwin->width], dwin->lines[k].chars.end(), ' ');
```
cspiegel, line 80: What's the rationale for changing `std::fill` here?  Is this a bug in MSVC that fails with the fill, or is it a bug in Gargoyle that just so happens to work with libstdc++?

awlck: The Microsoft STL does not like `std::fill` with one pointer and one `std::array<>::iterator`. [cppreference suggests](https://en.cppreference.com/w/cpp/algorithm/fill) that the first and second arguments to `fill` should be the same type, so I'm leaning towards libstdc++ being more lenient than the MS STL.

For reference, this is the compile error:

```
C:\Users\Adrian\Documents\Development\Cpp\garglk\garglk\wingrid.cpp(80,9): error: no matching function for call to 'fill'
   80 |         std::fill(&dwin->lines[k].chars[dwin->width], dwin->lines[k].chars.end(), ' ');
      |         ^~~~~~~~~
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.37.32822\include\xutility(5032,19): note: candidate template ignored: deduced conflicting types for parameter '_FwdIt' ('unsigned int *' vs. 'iterator' (aka '_Array_iterator<unsigned int, 256ULL>'))
 5032 | _CONSTEXPR20 void fill(const _FwdIt _First, const _FwdIt _Last, const _Ty& _Val) {
      |                   ^
```

garglk/cheapglk/cgfref.cpp
awlck: Yeah, this one is plain wrong. `02` would check for writing rather than reading.

It's also redundant with the definition I later added to my minimal `unistd.h` so it ought to go anyways.

cspiegel:
> At least TADS, Alan, and Hugo have upstream repositories who would presumably appreciate these changes.

Good call, and I also forgot that JACL has a resurrected repository at https://github.com/DangarStu/JACL. It's not super active, but Stuart was quick to merge a few pull requests I made last year.

terps/CMakeLists.txt
awlck: True, MSVC has a separate compiler option (`/Zc:__STDC__`) that enables this definition. But we don't *want* to have that defined, because doing so would hide away all the POSIX functions.

(MS takes the stance that, since POSIX functions aren't specified by ISO C, they should count as vendor extensions and thus be prefixed with underscores. Which is technically correct, I guess, but very annoying.)

cspiegel:
As an addition query, but not needed for this pull request, do you have experience running Windows builds with GitHub Actions?  Once this stabilizes, it'd be great to add a Windows runner to ensure we don't break the build.

garglk/wingrid.cpp
awlck: I got curious and looked it up, and libstdc++ really just `typedef`s a pointer to be the iterator, hence the original code worked there. The MS STL has a distinct iterator type, so it breaks. Same goes for comparing pointers to iterators.

Anyways, your adjusted `std::fill` call works.

awlck:
I don't really have experience running anything on GitHub Actions, I'm afraid. I never saw much of a reason for migrating away from AppVeyor.

But yes, figuring that out is on my to-do list. They have Windows images with Visual Studio, LLVM, and vcpkg preinstalled, how hard can it be? (Famous last words)

cspiegel:
I managed to get this up and running on Windows.  In addition to following the instructions in INSTALL.md, I had to set up a Visual Studio environment (I'm using Cygwin, but I'm simulating loading "x64 Native Tools Command Prompt", which updates $PATH with VS tools) in order to get access to mt.exe, from what I can tell.

But I can also build _without_ the environment if I pass `-DCMAKE_MT=llvm-mt` when running cmake.  It looks like this is building entirely with Clang (which I installed from the LLVM project directly), and not relying on Visual Studio.  If this is accurate, it seems that Visual Studio isn't a requirement for building, at least not at the moment: you just need clang, Qt, and vcpkg.

I admit to having no experience on Windows, so I may be incorrect.  And, of course, the whole point of this is .NET for Frankendrift, so that may require the full VS?  On the other hand, it wouldn't be unreasonable to be able to just download a self-contained .NET compiler.

Anyway, no matter what, I can at least build this on Windows now.

awlck:
I guess you might not strictly need the full VS install, yes. But I'm pretty sure you'd need the [Windows SDK](https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/), which contains the development files for the Win32 components, Microsoft CRT, etc. I'll have to try that out in a VM, though.

terps/tads/glk/osportable.cc
```diff
@@ -961,8 +1012,8 @@ os_is_special_file( const char* fname )
     // "..".  (We use OSPATHCHAR instead of '/' though.)
     const char selfWithSep[3] = {'.', OSPATHCHAR, '\0'};
     const char parentWithSep[4] = {'.', '.', OSPATHCHAR, '\0'};
-    if ((strcmp(fname, ".") == 0) or (strcmp(fname, selfWithSep) == 0)) return OS_SPECFILE_SELF;
-    if ((strcmp(fname, "..") == 0) or (strcmp(fname, parentWithSep) == 0)) return OS_SPECFILE_PARENT;
+    if ((strcmp(fname, ".") == 0) || (strcmp(fname, selfWithSep) == 0)) return OS_SPECFILE_SELF;
```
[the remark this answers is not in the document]
cspiegel: @awlck The original code builds fine for me with Clang on Windows; can this be reverted so that it deviates from upstream as little as possible?

With that done, I think we're in a good place to get this merged.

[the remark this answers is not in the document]
awlck: I've finally gotten around to changing this back. Visual Studio is not happy about it, but it does indeed build with Clang.

cspiegel:
Thanks for doing this!

PR #820, opened 2024-02-25
Allow resources to be loaded from memory

cspiegel:
This allows both Hugo and Adrift games to load media without the clumsy method of writing out SND and PIC files.


curiousdannii:
I'll want to add something like this to Emglken eventually too. Do you think this is a better method than a fake blorb, which is what I'd been thinking before?

terps/hugo/heglk/hemedia.c
```diff
@@ -51,6 +51,60 @@ void hugo_setgametitle(char *t)
 static schanid_t mchannel = NULL;
 static schanid_t schannel = NULL;
 
+#ifdef GARGLK
+struct GlkResource {
+  long offset;
+  int id;
+};
+
+static struct GlkResource glk_resources[8192];
```
curiousdannii, line 60: Why such a large fixed-length array?

cspiegel:
> I'll want to add something like this to Emglken eventually too. Do you think this is a better approach than a fake blorb, which is what I'd been thinking of doing before?

I think the fake Blorb approach is cleaner in one sense, as it requires no Glk extensions, although since at least two interpreters would need it, you'd probably want to factor out the "create Blorb file" code and stick it somewhere.

The other issue with a fake Blorb, at least for Hugo and SCARE, is that the resources are loaded on demand, and you don't know ahead of time where they are, so you'd have to create the Blorb file anew each time a new resource is loaded, which means caching everything that you've found, in the interpreter.

So, in short, I'm not sure which way is better.  This way keeps the interpreters trimmed down, but it requires extending Glk, which I'm not a huge fan of.

terps/hugo/heglk/hemedia.c
cspiegel: > Why such a large fixed-length array?

Because I hate dynamic array management in C but didn't want to have my refusal to implement one be the source of media-loading failure... but you're right, of course, that there won't be a game out there even _approaching_ this many resources.  Unless I convert this over to C++ I really don't want to do anything but a fixed-size array; what do you think would be a reasonable "max resources" value?  The old Hugo code had 1024, and even that's probably much larger than it needs to be.

curiousdannii:
I was thinking about this overnight, and I think this probably would be a better way than a fake blorb, unless the fake blorb is completely constructed ahead of a time (before even starting Glk). For Remglk/Parchment it could be quite messy to organise a fake blorb with the (potentially remote) GlkOte client, whereas it's easy to just send an arbitrary image by URL. Using resources from a Blorb is more efficient, but if there is no Blorb then sending by URL would be better.

terps/hugo/heglk/hemedia.c
curiousdannii: A few idea I can think of:

1. Use realloc to resize when needed
2. Implement an actual map in C++, with some helper functions, so that the C terps don't need to handle the map themselves
3. Use the memory pointer as the ID, calling a new Glk function that takes a pointer instead of an image ID.

I think I might like 3 the most. But it would mean keeping the resources files in memory permanently, while it looks like Scare reads the file each time you reference a new image. So that's a change I'd want to make anyway.

On the other hand terps that can load multiple resource files would then need a map for filenames, so they'd still need something... But at least such a map probably wouldn't need to be longer than a dozen entries. (Or are there games with even more resource files than that??)

terps/scare/os_glk.c
```diff
@@ -1737,6 +1737,85 @@ os_stop_sound (void)
...
+ * Use the Gargoyle-specific garglk_add_image_resource() to load images from memory.
+ */
+void
+os_show_graphic (const sc_char *filepath, sc_int offset, sc_int length)
+{
+  FILE *in = NULL;
+  unsigned char *data = NULL;
+
+  for (size_t i = 0; i < glk_images_count; i++) {
+    if (glk_images[i].offset == offset) {
```
curiousdannii, line 1763: What happens if multiple files have an image at the same offset? Or is Scare not one of the terps that can have multiple resource files?

cspiegel:
On balance I do think the API extension is the better approach, so I agree with your assessment.

Do you have any feedback on the proposed API functions?  I thought about allowing the caller to provide the resource ID (since in the case of Hugo and SCARE it'd be trivial, given that resources are uniquely identified by an offset into the file), but it just felt messier that way, although I guess in either case the interpreter would have to keep track of _something_.

You could also just have a single "add_resource" method that takes an enum and shunts things off to the appropriate sound/image backend, but that seemed unnecessary since it's just two functions.

curiousdannii:
Did you see my comment here? https://github.com/garglk/garglk/pull/820#discussion_r1501902450

The ideal API would depend a lot on whether or not multiple resource files is an actual issue or not. I thought it was a possibility, but can't remember for which formats it's an actual option. We could even just add a function like this:

```
glk_image_draw_from_file(winid_t win, char *filename, glui32 offset, glsi32 val1, glsi32 val2);
```

Then any caching of files and images would be entirely the Glk library's responsibility.

terps/scare/os_glk.c
cspiegel: That's an excellent point that I never considered.

Based on the earlier code, it looks like this is always looked up in the main game file. On line 3560:

```
#ifdef LINUX_GRAPHICS
  /* Note the path to the game file for graphics extraction. */
  gsclinux_game_file = argv[argv_index];
#endif
```
So this is safe.

But SCARE _does_ seem to be able to use different resource files!  So you're absolutely right, the equivalent code in the SCARE interpreter is not correct.  I'll get that fixed up.  Nice find!

curiousdannii: This comment was on Scare 😉

cspiegel: Sorry, I mean _Hugo_ can have different resource files, not SCARE!

curiousdannii:
Kind of unrelated, but I suspect for the future of Emglken it may be best if we can move the terps to using Glk file streams rather than reading files directly through libc, so any new APIs that wouldn't require the terp to read a file itself would be better from my perspective.

cspiegel:
> Did you see my comment here? [#820 (comment)](https://github.com/garglk/garglk/pull/820#discussion_r1501902450)
> 
> What the ideal API would look like depends a lot on whether or not multiple resource files is an actual issue or not. I thought it was a possibility, but can't remember for which formats it's an actual option.
> 
> We could even just add a function like this:
> 
> ```
> glk_image_draw_from_file(winid_t win, char *filename, glui32 offset, glsi32 val1, glsi32 val2);
> ```
> 
> Then any caching of files and images would be entirely the Glk library's responsibility.

Just saw the comment now (github didn't refresh things for me).

It does look like Hugo, at least, supports multiple files.  The game "Trading Punches", for example, has the files `TRADGFX` and `TRADMUS`, and the code to load resources takes a file handle instead of a filename.

I would say that that makes it hard to implement `glk_image_draw_from_file`, and that's true, but to fix image loading as I have it would require the filename, too, to form a unique key of filename + offset, so that's moot.

With that said, we'd also need to add a similar function for playing sound, and maybe a scaled version of the graphics.  The upside to the "add" functions is that they allow all the existing media handling code to be used, although this proposed function would greatly simplify things from the _caller's_ perspective, which is probably the most important thing.

One downside to this proposed function is that it won't allow resource loading from game formats that have to decode/deobfuscate their resources, although this is only a theoretical concern, since the two formats known to need this (Hugo + Adrift 4) don't do that.

curiousdannii:
The number of functions taking resource IDs is medium, so we *could* add duplicate _from_file variants from each. But maybe better not to.

Alternatively, we could have a `glk_add_resource_from_file(char *filename, glui32 offset, glui32 len)` function, which could be designed on the Glk library side to be quite cheap, so that we could then call it inline `glk_image_draw(win, glk_add_resource_from_file(filename, offset, len), val1, val2)` and not need to have a map in the C terp. Likewise your original `garglk_add_image_resource` function could be designed to be cheap enough to include inline like that. It would just borrow the data, not take ownership, so if the terp needed to do any decoding that could work too.

Rather than separate functions for images and sounds they could have a parameter indicating the resource type?

cspiegel:
I've been experimenting and here's where things stand:

I like the idea of `glk_add_resource_from_file`.  By doing this, all the file loading code is consolidated into the Glk library, instead of duplicating it in the interpreters.  As it stands, the API function I've implemented is this:

```
glui32 garglk_add_resource_from_file(glui32 type, const char *filename, glui32 offset, glui32 len, glui32 *id);
```

This returns true on success, storing the ID in `*id`, false otherwise. This is necessary since there might be issues loading/reading the file. "type" is either `giblorb_ID_Pict`, or `giblorb_ID_Snd`, which means the Blorb layer needs to exist, but that's probably not a huge deal: if you don't have Blorb support you probably don't have multimedia support in general.

It really is better to have a function that creates a resource ID that can be used by existing functions. It just makes things cleaner.

As for the naming, I think it's better to have names that aren't part of the "official" Glk or Blorb namespace, since that's Zarf's province. But it wouldn't be a bad thing to have an unofficial namespace for things that aren't _just_ Gargoyle, e.g. `glkext_add_resource_from_file` or something.  If it ever got absorbed into Glk, then it's just a matter of adding a macro or wrapper function (tagged as deprecated) to shunt things to the new function.

I've implemented this as a proof of conecpt in #824 if you want to take a look.

cspiegel:
Closing in favor of #825.

PR #825, opened 2024-02-29
Add garglk_add_resource_from_file() API function

cspiegel:


garglk/garglk.cpp
```diff
@@ -115,3 +128,99 @@ void garglk_window_get_size_pixels(window_t *win, glui32 *width, glui32 *height)
...
+
+glui32 garglk_add_resource_from_file(glui32 usage, const char *filename_, glui32 offset, glui32 len)
+{
+    std::string filename(filename_);
+
+#ifdef _WIN32
+    static const std::vector<std::string> reserved_names = {
+        "con", "prn", "aux", "nul", "com1", "com2", "com3", "com4",
+        "com5", "com6", "com7", "com8", "com9", "lpt1", "lpt2",
+        "lpt3", "lpt4", "lpt5", "lpt6", "lpt7", "lpt8", "lpt9",
```
curiousdannii, line 166: Are these needed now that there's a restriction that the file has to be in the same folder as the storyfile?

cspiegel: Windows is bizarre and treats these names as special regardless of where they are, at least some of the time.  I wrote a small test program and Windows 10 lets you open an absolute path that ends in "\CON", and treats it like stdin (Windows 11, in my testing, only does this for the filename "CON", not a full path ending in "\CON").

"CON" is a problem because reading from it will hang the program if stdin is connected to anything that will wait for input, and I have no real insight into how various versions of Windows treat the _other_ files for reading, so it's safest to just ban them all.

curiousdannii: Oh yeah, I have vague memories of seeing that before.

https://superuser.com/questions/86999/why-cant-i-name-a-folder-or-file-con-in-windows

Weird that it also applies to files of those names with any extensions!

cspiegel:
I'll merge this in now; if there are tweaks that need to be made in the future as issues arise, that's fine.  We can treat this as a proof of concept and adjust as necessary.

PR #833, opened 2024-03-28
Add and check a macro protecting garglk_tads_os_banner_size()

cspiegel:
This needs to be submitted to upstream, but upstream can't assume that Gargoyle is the current version; so restore the old invasive way, using the new TADS-specific API function only if its (new) associated macro exists.

### DIFF
--- a/terps/tads/glk/osglkban.c
+++ b/terps/tads/glk/osglkban.c
@@ -28,6 +28,10 @@
 #include "os.h"
 #include "glk.h"
 
+#ifndef GLK_MODULE_GARGLK_TADS_OS_BANNER_SIZE
+#include "garglk.h"                 /* for-size to-contents hack */
+#endif
+
 typedef struct os_banner_s *osbanid_t;
 typedef struct banner_contents_s *contentid_t;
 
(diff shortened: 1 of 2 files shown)


terps/tads/glk/osglkban.c
```diff
@@ -28,6 +28,10 @@
 #include "os.h"
 #include "glk.h"
 
+#ifndef GLK_MODULE_GARGLK_TADS_OS_BANNER_SIZE
```
curiousdannii, line 31: Wouldn't it be simplest to just check for `GARGLK`?

curiousdannii: Haha, just noticed that was literally what was removed in this commit: https://github.com/garglk/garglk/commit/8a9998a3a8e478b477853a6d0d3ab5518c72ac6a

cspiegel: You're right that this is incomplete (not sure what I was thinking).

This needs to be in `#ifdef GARGLK`, since it includes `garglk.h`, but then also "needs" the check for the banner size feature, since there exist Gargoyles out there which don't support it.

Really, that's probably unnecessarily careful: compatibility with older Gargoyles really isn't useful.  After all, they include their own copy of TADS, and if somebody really feels the need to update TADS in an older Gargoyle release, they can bring over this older method manually.  It'd probably be more practical to just "know" that Gargoyle provides this API function.  I was trying to be completely proper here, but in all honesty it's unnecessary.

cspiegel: With that in mind, this pull request becomes superfluous.  I'll now see about integrating Gargoyle differences into the tads-runner repository so we can hopefully have complete synchronization between the two.

cspiegel:
As noted, this is unnecessary.

PR #845, opened 2024-05-24
Give access to Gargoyle color settings

cspiegel:


curiousdannii:
Is this just a convenience over `glk_style_measure` that lets you get both at once?

cspiegel:
It is, but only because I _totally_ forgot that `glk_style_measure` exists! There's no need for this function; thanks for pointing it out. I'll just revert this change.

PR #851, opened 2024-06-18
a Hugo parser Engine fix

roodyyogurt:
Added some lines in heparse.c that fix a bug that comes up if a player is trying to refer to a known out of scope object when there is another known out of scope object that shares an adjective.  Without this fix, the engine locks onto the first defined object, ignoring nouns, and always sending that first object to ParseError.

This is my first time doing a pull request so if I'm doing something wrong, someone probably has my e-mail somewhere and can contact me directly.

### DIFF
--- a/terps/hugo/source/heparse.c
+++ b/terps/hugo/source/heparse.c
@@ -1087,7 +1087,12 @@ int MatchObject(int *wordnum)
 
 				/* definitely not this object */
 				else
+				{
 					SubtractPossibleObject(i);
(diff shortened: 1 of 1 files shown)


cspiegel:
I sync Gargoyle with upstream interpreter releases, so your fix there would get pulled in here as well on sync, but that assumes that upstream makes a release before the next Gargoyle; and there's no harm in the change being here, since it won't conflict with the upstream release, so I'll merge this in to ensure we have a fix in the next release.  Thanks!

PR #853, opened 2024-07-14
Patch for Transporter

angstsmurf:
See discussion here:
https://intfiction.org/t/unsolved-games/67447/38

This game tries to read a property from an object without checking if it is valid in a routine at 0x4bd1. This will be called with a negative value when trying to ASK the wizard about things, and with the number typed when trying to TYPE a number, which may be larger than the number of objects.

We fix this by checking if the argument is a valid object (removing some redundant code to make room).

This will be the largest patch in Bocfel. Possibly there is a simpler solution.

Original code:
```
 4bd1:  41 01 00 00 03          JE              L00,#00 [FALSE] 4bd7
 4bd6:  b1                      RFALSE          
 4bd7:  52 01 01 03             GET_PROP_ADDR   L00,#01 -> L02
 4bdb:  52 01 01 00             GET_PROP_ADDR   L00,#01 -> -(SP)
 4bdf:  2d ff 00                STORE           Gef,(SP)+
 4be2:  a0 ff c5                JZ              Gef [TRUE] 4be8
 4be5:  a4 ff ff                GET_PROP_LEN    Gef -> Gef
 4be8:  e8 bf ff                PUSH            Gef
 4beb:  57 00 02 00             DIV             (SP)+,#02 -> -(SP)

 (The rest of the routine, unchanged from here on:)
 4bef:  55 00 01 04             SUB             (SP)+,#01 -> L03
 4bf3:  0d 05 00                STORE           L04,#00 
 4bf6:  63 05 04 80 11          JG              L04,L03 [TRUE] 4c0a
 4bfb:  6f 03 05 00             LOADW           L02,L04 -> -(SP)
 4bff:  61 02 00 00 03          JE              L01,(SP)+ [FALSE] 4c05
 4c04:  b0                      RTRUE           
 4c05:  95 05                   INC             L04
 4c07:  8c ff ee                JUMP            4bf6
 4c0a:  b1                      RFALSE   
```
Patched code:
```
 4bd1:  42 01 01 80 09          JL              L00,#01 [TRUE] 4bdd
 4bd6:  c3 8f 01 02 31 00 03    JG              L00,#231 [FALSE] 4bde 
 4bdd:  b1                      RFALSE          
 4bde:  52 01 01 03             GET_PROP_ADDR   L00,#01 -> L02
 4be2:  2d ff 03                STORE           Gef,L02
 4be5:  a0 ff c5                JZ              Gef [TRUE] 4beb
 4be8:  a4 ff ff                GET_PROP_LEN    Gef -> Gef
 4beb:  57 ff 02 00             DIV             Gef,#02 -> -(SP)
```


cspiegel:
Let me think about how to deal with this... I'm not sure how broad patching should be, and maybe I set a bad precedent with "Robot Finds Kitten".  Meaning that any game can easily include bad code that breaks, and where should the line be drawn on what to fix?  I'm always going to include Infocom fixes for what I think are obvious reasons, but I half wonder if maybe I should extend Bocfel to load patches from an external file, and then we can create a repository of fixes for non-Infocom games.  That's something that could then be distributed with Gargoyle, Spatterlight, etc, so it can be kept up to date and have whatever fixes we want in there.  It'd also allow on-demand patch updates similar to virus definitions.  I probably won't go that far in Gargoyle, but it would at least be possible.

Bocfel already can load patches from `bocfelrc`, so it'd be pretty easy to extend it to load from a dedicated file that the user doesn't ever need to worry about.  With the current setup, using existing parsing code, it'd look something like:

```
# Transporter tries to read a property of non-existent objects, so we add a bounds check.
[1-960729-1ac6]
0x4bd1 28 [41 01 00 00 03 b1 52 01 01 03 52 01 01 00 2d ff 00 a0 ff c5 a4 ff ff e8 bf ff 57 00] [42 01 01 80 09 c3 8f 01 02 31 00 03 b1 52 01 01 03 2d ff 03 a0 ff c5 a4 ff ff 57 ff]
```
Which isn't pretty, really, but doesn't need to be.

The downside, of course, is that it's divorced from Bocfel itself, but in general I think that's not a horrible thing; most users are going to be using it from another source, e.g. Spatterlight or Gargoyle, which can ensure that it's up to date, and as mentioned, can do it more quickly when necessary.  And the upside is that it's something other interpreters could make use of if they wanted to.

What are your thoughts on this?

angstsmurf:
Having an external file with patches sounds reasonable to me. As long as there is an official patches file that is always included, I guess it might simplify the code to have it includes all the Bocfel patches, including the Infocom ones.

Closing this for now.

PR #857, opened 2024-08-06
Update Bocfel to 2.2

cspiegel:


curiousdannii:
Exciting to see the Z6 support!

Does this also include the no-stdio changes you were working on? If so I'll give it a test to see if I can get it to build and run with Remglk-rs.

cspiegel:
It doesn't have the no-stdio stuff yet, no; I really haven't had the chance to properly test it. That said, I'm attaching a diff against Bocfel 2.2 which implements it. If you're willing to test this and let me know how it works, I'll get it merged in and into a point release in short order, so it can be made official.

[no-stdio.diff.gz](https://github.com/user-attachments/files/16511218/no-stdio.diff.gz)

curiousdannii:
I'll try it, but don't wait for me before merging this.

PR #865, opened 2024-08-24
Update Bocfel to 2.2.1

cspiegel:


terps/bocfel/io.cpp
```diff
@@ -35,6 +35,20 @@ extern "C" {
 #include "unicode.h"
 #include "util.h"
 
+#ifdef ZTERP_NO_STDIO
+#ifndef ZTERP_GLK_UNIX
+#error ZTERP_NO_STDIO requires a Unix Glk
+#endif
+
+#include <glkstart.h>
```
curiousdannii, line 43: Should this also be wrapped in `extern "C"`? I'll add the extern C wrapping to my remglk-rs library, but it couldn't be good to wrap this too, just in case, unless double wrapping is bad?

curiousdannii:
So the no-stdio is working up until I try to save/restore. It says that the savefile it created is corrupt. I'll attach it in case you can quickly see what is wrong, and I'll keep digging to see if I can work out what's happening.
[advent save.zip](https://github.com/user-attachments/files/17198366/advent.save.zip)

terps/bocfel/io.cpp
cspiegel: Yep, nice catch.  I was building against (an experimental branch of) Gargoyle which sets `extern "C"` in the header, so I never noticed this.  I'll have it added in the next release.

cspiegel:
> Hmm. I suspect the problem is caused by Bocfel seeking past the end of the file. Libc just expands the file, but RemGlk-rs takes the Glk spec literally, and clamps the position to the current file:

Well, that's unfortunate!  It looks like cheapglk (and by extension Gargoyle + probably Zarf's other Glks, and probably most other Unix Glks) just passes along the info to `fseek()` which, as you note, is perfectly legal on Unix. And Windows Glk looks to just call `fseek()` as well and, from what I can find, seeking beyond the end of file on Windows is also valid. So it works by happenstance, due to the "obvious" seek implementation supporting it incidentally on almost all platforms.

That being said, unless it's determined to be a defect in the Glk standard, Bocfel really shouldn't be trying to seek beyond the end.  Looking at the code, it seems there's just one place(?) that is causing this issue, in `stack.cpp`'s `write_chunk()`:

```
io.seek(8, IO::SeekFrom::Current); // skip past type and size
```

Is that the source of it?  If so, that's easy enough to fix to be compatible with Glk:

```
// Type and size, to be filled in below.
io.write32(0);
io.write32(0);
```

curiousdannii:
I really doubt Zarf would decide to patch his libraries to block this, he's much more likely to just amend the Glk spec to match what everyone is doing. It's up to you if you want to bother making Bocfel spec compliant, but I wouldn't at this time.

cspiegel:
It does look like it's just that one place in the code, and since stdio isn't technically required to allow seeking beyond the end, either, I'll just make the change (which is not an argument one way or the other for what Glk should do).  If it were more pervasive in Bocfel I'd probably just leave it, but this is a trivial change.

PR #871, opened 2024-09-04
FrankenDrift integration

awlck:
This PR introduces the FrankenDrift interpreter into the codebase, and adjusts the Windows and Mac build scripts to include it. (On the Mac, I switched to building libgarglk as a dynamic library, since that happens to be way more convenient for the FrankenDrift build process. Let me know if this is not acceptable.)

Work still needs to be done on the Linux build scripts. I am somewhat out of my depth there.


cspiegel:
No problem with a shared libgarglk on Mac.  I disabled defaulting to shared libraries because, on Unix-y systems it's effectively pointless for Gargoyle and adds one more potential point of failure; but on a self-contained bundle like Mac (or Windows, for that matter), it's not a problem, and maybe is even a slight advantage, since you're saving a bit of disk space.

I'll take a look at this and see about integrating it into the Linux builds.

cspiegel:
I've just got a small pull request at awlck/garglk#1 to match the code style (tabs to spaces) and split off a comment to explain a piece of code that "looks" wrong. But nothing substantive.

I was able to build this with no problems on Linux.  It produces a self-contained binary with the only difference to other interpreters being that it dynamically loads libgarglk.so, _but_ there is no effective difference in that, since it'll use the same search path.

The one "oddity" is that the build of FrankenDrift is not integrated into the CMake build at all; so it's possible to tell CMake you're doing a FrankenDrift build and then just not actually go through with it.

What do you think of adding a step in the build process to run the FrankenDrift builder via CMake?  It'd need to be a bit special-cased, to copy to `build/dist` for those that use it (Windows, Mac, Linux AppImage, I think); and the "normal" Unix install would need to learn about it; but it should make it harder to mess up the build, and will make it easier for distributors who use the normal cmake→build→install procedure.

cspiegel:
See awlck/garglk#2 for a start to CMake integration.

It'll need to be expanded to all operating systems/build types, but this does work for a basic Unix build right now.

awlck:
Alright, then.

* FrankenDrift gets build through CMake (where it's supported)
* FrankenDrift gets cross-compiled to arm64 if that's what is set up in CMake. (Although I don't have an arm64 Windows device to test this with.)
* FrankenDrift's linkage of libgarglk has been untangled, and AoT dynamic linking is used on all platforms.

The only remaining fly in the soup (that we also can't do anything about) is that .NET 8 on the Mac only supports macOS 12 and newer.

cspiegel:
Thanks!  As for ARM builds, I've never actually tested them either, for the same reason.  I assume if somebody tries and they fail they'll provide a bug report.

No real harm on the Mac side of things: the majority of users should be on new enough systems to run a Monterey-built binary (which is where I build the "main" Mac release). Those on older systems just won't get Frankendrift.

PR #873, opened 2024-09-07
Sync with upstream TADS

cspiegel:
@awlck You can test your monospace fixes here to ensure things are properly integrated with Gargoyle.


curiousdannii:
FYI, I'd really like to try to get colour support into TADS by the next Gargoyle release. I can make that more of a priority of you're wanting to do a release soon.

cspiegel:
I'm hoping to get a release out pretty soon. I was at least going to wait till the comp is over, but I'm fine pushing it to give time for your color additions. There's already going to be Adrift 5 and Infocom graphics support, so it'd be great to have an improved TADS interpreter as well, making for a pretty significant release.

curiousdannii:
Probably easiest for me to do the updates in Gargoyle and then bring them to the tads-runner repo afterwards. I'll work on it the week of the 16th.

It would be good the sync the Hugo changes too.

awlck:
> You can test your monospace fixes here

Working as expected!

awlck:
> I'd really like to try to get colour support into TADS by the next Gargoyle release

TADS 3 is simple enough -- the built-in HTML parser already recognizes `<font color=...`, and it's just a matter of populating the `os_set_text_color` function (see attached patch). TADS 2 is more involved, since *its* built-in HTML parser needs to be amended to recognize it.

[t3-color.patch](https://github.com/user-attachments/files/16923172/t3-color.patch)

(I kinda noticed this back when I did the monospace patch but never got around to finishing it and submitting a PR... 😅 probably got myself confused on T2's HTML parser.)

curiousdannii:
Thanks, that's helpful. Though the banners colour code is more involved.

curiousdannii:
So I've started looking into this, and it's not looking good.

First up is that TADS 2 just isn't designed to do colours unless in HTML mode:

https://github.com/tads-intfic/tads-runner/blob/620f8cd6c8b606812cf8511b8b8183b8e4282d06/tads2/bif.c#L3282-L3293

So I thought, okay, I eventually wanted to look at implementing HTML TADS for Glk. Let's just do it.

The [porting notes](https://www.tads.org/t3doc/doc/htmltads/porting.htm) explain what you need to do. You have to implement half a dozen classes, but it is designed to be portable across implementations.

But I'd never actually looked at those interfaces. And now I see that they are designed for *painting text at specific pixels*! Ironically, it would be almost impossible to actually implement HTML TADS in a browser, let alone Glk. (I guess with a browser maybe you could absolutely position `<div>`s rather than doing any actual flowing HTML?)

https://github.com/realnc/qtads/blob/966f43107aeba413e6122189b071fe0fa32dc4fd/htmltads/htmlsys.h#L1175-L1182

So unfortunately full HTML TADS support is just never going to be feasible.

But Adrian previously showed that supporting more HTML is possible in the HTML-to-text mode converter (https://github.com/tads-intfic/tads-runner/pull/27). But what it means is that any future formatting support we add won't be through clean implementations of TADS' interfaces, but instead through edits deep in the VMs. Oh well, so be it.

awlck:
> So I thought, okay, I eventually wanted to look at implementing HTML TADS for Glk. Let's just do it.
> 
> The [porting notes](https://www.tads.org/t3doc/doc/htmltads/porting.htm) explain what you need to do. You have to implement half a dozen classes, but it is designed to be portable across implementations.
> 
> But I'd never actually looked at those interfaces. And now I see that they are designed for painting text at specific pixels!

I don't think you *have to* use these interfaces to implement HTML formatting support. They form the foundation of the Windows HTML TADS interpreter, as well as QTads, presumably because this is a convenient model for stand-alone desktop apps that *don't* go through another layer of abstraction (like a browser or Glk library).

As I understand it, raw HTML markup passes through the `os_print`/`os_printz` functions declared in `tads2/osifc.h`. (The portable HTML TADS code implements them here: https://github.com/realnc/qtads/blob/966f43107aeba413e6122189b071fe0fa32dc4fd/htmltads/oshtml.cpp#L616) So it should be possible to start from those, either handing the HTML straight to a browser, or otherwise implementing our own HTML parser for Glk targets.

curiousdannii:
> So it should be possible to start from those, either handing the HTML straight to a browser, or otherwise implementing our own HTML parser for Glk targets.

Expanding the existing HTML->text mode parsers seems the easiest way to go about that.

cspiegel:
Sounds like it'll be more work than can be done for a soon-ish release, unfortunately?

I'm still planning on waiting till after the comp to do a release, so that's a month; if some TADS-related changes get in before then, great, but if not, we can punt.

curiousdannii:
I made a start on expanding the parser, but I didn't get it finished yet. If you want to make an end of the year Gargoyle release, don't wait for me. Sorry!

PR #875, opened 2024-09-08
Sync with upstream Hugo

cspiegel:
There hasn't been an "official" Hugo release since 2021, but the repository is effectively in maintenance mode, so current master just has bug fixes, which are worth bringing over.


curiousdannii:
Have your media changes been sent upstream yet?

cspiegel:
Not yet; I'm not sure if I'm making some Gargoyle-specific assumptions.  I'll take a look at the changes and clean them up for inclusion with upstream.

PR #881, opened 2024-09-10
Update Level9 terp version on CMakeLists.txt

cibersheep:
Just a cosmetic change

### DIFF
--- a/terps/CMakeLists.txt
+++ b/terps/CMakeLists.txt
@@ -295,8 +295,8 @@ endif()
 #
 # Level9 5.1
 #
-# The bitmap graphics from the US Macintosh release of Lancelot
-# now display correctly.
+# Disabled the status bar that just contained a static version string.
+# Added graphics support.
 #
 if(WITH_LEVEL9)
     terp(level9
--- a/terps/CMakeLists.txt
+++ b/terps/CMakeLists.txt
@@ -293,10 +293,10 @@ endif()
 
 # ------------------------------------------------------------------------------
 #
-# Level9 4.1
+# Level9 5.1
 #
-# Disabled the status bar that just contained a static version string.
-# Added graphics support.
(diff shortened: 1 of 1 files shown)


terps/CMakeLists.txt
```diff
@@ -293,10 +293,10 @@ endif()
 
 # ------------------------------------------------------------------------------
 #
-# Level9 4.1
+# Level9 5.1
 #
-# Disabled the status bar that just contained a static version string.
-# Added graphics support.
+# The bitmap graphics from the US Macintosh release of Lancelot
+# now display correctly.
```
cspiegel, line 299: The comments here document the changes made to the interpreters for inclusion with Gargoyle, so this "should" remain as it was.

I say "should" because Level9's changes are more extensive than originally listed.  But more than that, listing the changes in the CMakeLists file is fragile, since you have to remember to update it (as you have to remember to update the version number, which I clearly didn't).

For upstream interpreters, I've begun putting the original versions in orphaned branches, so I can merge them into master, and easily keep a record via Git of which changes were made, obviating the need for the list here.

There are some significant changes from upstream Level9, and it's probably time to evaluate whether they're still correct. If you can restore the original comment here (but keep the 5.1 version change), I'll merge it then start the upstreaming process for Level9 5.1 properly. You can rewrite the commit and force push, or just make a new commit. Doesn't matter to me.

Eventually I'll just remove this record of differences, since Git makes it easy to see.  For example:

```
cd terps/glulxe
git diff upstream-glulxe -- .
```

cibersheep: Sorry about that, I should have asked before doing the MR ;)
With the last commit should get back the comment. Let me know if this is wrong

cspiegel: No problem! This lit a fire under me to get the Level9 stuff properly upstreamed, so that's good.

PR #884, opened 2024-09-12
Sync scare

cspiegel:


curiousdannii:
Where is the upstream for Scare now?

cspiegel:
Yep, this is 1.3.10 from the IF Archive.  There's no new functionality here: I'm just trying to get as many interpreters as I can in the "upstream-[foobar]" branches, to make it easier to keep track of all changes that Gargoyle has made.

The idea is that you can go into any interpreter directory and say something like this:

```
git diff upstream-[foobar] -- .
```
That'll show the difference between the "pristine" upstream sources and whatever is in the current branch.

It's especially useful for interpreters that are still maintained, because upgrading is basically just a merge, instead of trying to manually figure out the differences. But even for something like Scare, it's good to be able to quickly see what's been changed.

PR #886, opened 2024-09-13
Add different cases for compiling for Garglk and Linux

### LINKED ISSUE #883: Couldn't build terp/magnetic 2.3.1 under Ubuntu 20.04
https://github.com/garglk/garglk/blob/a682063f542cdf63388af166845d5e42189a9b44/terps/magnetic/Glk/glk.c#L6217C8-L6217C14

I had an error while building this terp even adding MACROS GARGLK to the CMakeList. Now I see there's because of the linkage section mentioned above.
I know this is very specific to me (under Linux compiling for Gelek and not for Garglk), but would it be fine if I split the current GARGLK cases into two? Something like TARGET_OS_UNIX, and GARGLK for the different cases?

cibersheep:
This should enable compiling the terp for Linux systems (I tested againt remglk using macros MAGNETIC_GLKUNIX) and for Garglk (with macros MAGNETIC_GLKUNIX and GARGLK).

But this needs to be tested as I may slip something for Garglk.
Also I'm unsure about the hint system (defs.h line 353)

It should fix #883


terps/magnetic/Generic/defs.h
```diff
@@ -35,7 +35,7 @@
 *       correct number of bits on your system !!!
 \*****************************************************************************/
 
-#ifdef GARGLK
+#ifdef MAGNETIC_GLKUNIX
```
cspiegel, line 38: OK, so there's a wrinkle here... `GARGLK` really means _three_ things, unfortunately:

1. Unix Glk startup (fixed in this request)
2. Running under Gargoyle
3. Changes from upstream

The only thing it _should_ be used for is 2. But because of 3, if you don't define `GARGLK`, then you'll get bad code (e.g. loads of pointer sign comparison warnings, which are technically invalid code even if most/all compilers do the "expected" thing; plus possibly passing the wrong types to sprintf in one instance).

I made these changes to Magnetic _before_ I started the whole upstream branch thing, so I used `#ifdef GARGLK` to indicate my changes.  Given how things work now, it's better to just make the changes, and use `git diff` to see the differences.

So what I'd ask you to do is this: find all `GARGLK`-wrapped code which is clearly a bug fix.  To my recollection this is a call to sprintf() in the Glk code, and various casts/type-related stuff in the Glk code as well as the generic code.

Just completely get rid of the code that isn't protected by the macro.  For example:

```
#ifdef GARGLK
type8s* hint_contents = 0;
#else
type8* hint_contents = 0;
#endif
```
becomes
```
type8s* hint_contents = 0;
```
And do that here in `defs.h`, too: let's just rip out the old typedefs.  `stdint.h` is from C99, so we can unconditionally use it. If somebody is still using a C89 compiler, it's on _them_ to get this working.

There is a bunch of graphics-related Gargoyle changes that can remain wrapped in `GARGLK` for now.  I haven't decided how to deal with that; but we really should just have bug fixes in the code here, unconditionally.

cibersheep: «No so fast Joan» :)
Thanks for taking the time to explain. I'll take a look into it, but it has to be in a couple of days (I go on a trip).

cspiegel:
I noticed that both Level9 and Agility have similar Glk startup code, so I'd expect any issues that Magnetic gives you, these would too.. are they working?  What was the error you were getting with Magnetic?  Maybe there's a deeper issue that needs to be resolved here, too.

cibersheep:
> Maybe there's a deeper issue that needs to be resolved here, too.

I may have addressed this wrong. I'll make some changes and propose them.
My issue was from all the GARGLK cases some requirements for Garglk doesn't work for other implementations as Gelek Vanilla (yet)

I think I need to split these cases. Are they specific for Garglk, right?

```
/terps/magnetic/Glk/glk.c:5963:14: warning: implicit declaration of function 'garglk_set_story_name' [-Wimplicit-function-declaration]
 5963 |       if (s) garglk_set_story_name(s+1);
      |              ^~~~~~~~~~~~~~~~~~~~~
[ 14%] Linking C executable magnetic
/usr/bin/ld: CMakeFiles/magnetic.dir/magnetic/Glk/glk.c.o: in function `gms_startup_code':
glk.c:(.text+0x7a42): undefined reference to `garglk_set_story_name'
/usr/bin/ld: glk.c:(.text+0x7a76): undefined reference to `garglk_set_story_name'
```

and 

```
/terps/magnetic/Glk/glk.c:6261:3: warning: implicit declaration of function 'garglk_set_program_name' [-Wimplicit-function-declaration]
 6261 |   garglk_set_program_name("Magnetic 2.3");
      |   ^~~~~~~~~~~~~~~~~~~~~~~
/home/cibersheep/Documents/UT-Projects/gelek-vanilla/build/x86_64-linux-gnu/terps/magnetic/Glk/glk.c:6262:3: warning: implicit declaration of function 'garglk_set_program_info' [-Wimplicit-function-declaration]
 6262 |   garglk_set_program_info(
      |   ^~~~~~~~~~~~~~~~~~~~~~~
[ 14%] Linking C executable magnetic
/usr/bin/ld: CMakeFiles/magnetic.dir/magnetic/Glk/glk.c.o: in function `glkunix_startup_code':
glk.c:(.text+0x7ffc): undefined reference to `garglk_set_program_name'
/usr/bin/ld: glk.c:(.text+0x800d): undefined reference to `garglk_set_program_info'
``

File: terps/magnetic/Generic/defs.h (hunk 1 above)
Comment (cibersheep) [reply terps/magnetic/Generic/defs.h]: If I hope I did this correctly and should compile for Garglk.

**Comment by cspiegel** [comment]
Yep, that's correct.

Gargoyle has always provided some Glk extensions, such as those you've highlighted.  It also defines the `GARGLK` macro, so the interpreters could check that to see if the extensions exist. In a sense `GARGLK` meant "these API extensions are available."

Since the interpreters were imported into Gargoye, and not intended to be used elsewhere, `GARGLK` also meant "running in Gargoyle", not just  "API extensions available".

But those are two separate things! It doesn't matter much when everything's running under Gargoyle, especially because it's all in one repo: if a change is needed, it can be done everywhere all at once.

Now that these interpreters are being used elsewhere, some of the assumptions are causing problems like the one you're seeing.  Since Gargoyle is a Unix Glk (regardless of actual platform it's running on), some interpreters assume `GARGLK` means Unix Glk. Though, as a side note, since Gelek-Vanilla uses remglk, it's fine assuming a Unix Glk, since that's what remglk is.

There are also, apparently, some changes that mesh better with Gargoyle, such as in Agility's `atexit` handler: if running under Gargoyle, it returns early, avoiding a call to `glk_exit` or writing to the main window.  The reason for this? Not sure (the commit message is "Added fixes to base Agility code", so that doesn't help).

Anyway, I think the right thing to do if you're a Glk implementation that is _not_ Gargoyle: don't define `GARGLK`.  Gargoyle uses its knowledge of its own implementation to make decisions on how to act which might not be valid under other Glk implementations. If the lack of `GARGLK` breaks another Glk implementation, we should modify the interpreter accordingly.  What that means depends on the specific situation.

Magnetic has that issue I mentioned where lots of the bug fixes are guarded by `GARGLK` and shouldn't be, so that's another thing to fix.

Basically.... build without `GARGLK` defined and fix what breaks.  It'll be really nice to have the interpreters be a bit more independent from Gargoyle.

File: terps/CMakeLists.txt [hunk 2]
```diff
@@ -327,7 +327,8 @@ endif()
 if(WITH_MAGNETIC)
     terp(magnetic
         SRCS magnetic/Generic/emu.c magnetic/Glk/glk.c
-        INCLUDE_DIRS magnetic/Generic)
+        INCLUDE_DIRS magnetic/Generic
+        MACROS MAGNETIC_GLKUNIX GARGLK)
```
cspiegel, line 331: This shouldn't define `GARGLK` since that's already defined in `glk.h`.

cspiegel:
With the one change mentioned regarding not defining `GARGLK`, I think this is fine to merge in; if you want/need to make changes to other interpreters, we can do it in separate pull requests.  So unless you have further changes here, I'll get this merged in after that small fix.

terps/CMakeLists.txt
cibersheep: Do you want me to update the version from 2.3 to 2.3.1 in the code? Or is it better to do it in a different request?

cibersheep:
> Basically.... build without `GARGLK` defined and fix what breaks. It'll be really nice to have the interpreters be a bit more independent from Gargoyle.

Oh, that's nice. I can try to do this at my own pace (as I already kind of doing it already)

cibersheep:
One more thing. Have you tried to build the terps for arm64?
I may need to double check the code, but it seems to load games in Gelek Vanilla for x86, but gives some errors loading the same games in arm64 (see https://gitlab.com/cibersheep/gelek-vanilla/-/issues/36 )

It may be worth to open an issue upstream as, if confirmed, may be a bug introduced between versions 2.3 and 2.3.1

cspiegel:
I hadn't actually tested arm64 as I don't have a setup that makes it easy, but I just installed a cross compiler and did a test against cheapglk, and I can reproduce it.  It's not an upstream bug, but a bug I introduced, and has to do with the fact that `char` is unsigned on aarch64.  I'll merge this, then add a fix for the bug.

PR #890, opened 2024-09-18
Revert type changes to upstream Magnetic

cspiegel:
The Magnetic code has loads of pointer type mismatches, essentially assuming that type8s (signed char) is compatible with char, which it isn't. char can be unsigned (and is, on aarch64), and even where it's signed, char and signed char are distinct types, and you cannot convert between pointers to the two without casting.

Originally I used "char" for "type8s" since it is generally used synonymously with "char" in the code, but there are indeed some areas where it _needs_ to be signed, breaking things on aarch64. While compilers are allowed to reject invalid pointer conversions between differently-signed integers of the same rank, in reality they don't. GCC doesn't even warn about such conversions by default, although the standard mandates a diagnostic.

Clang does warn, so for Clang (and GCC, for good measure), use #pragmas to disable this particular warning. Eventually if a strict compiler comes along we'll have to just litter the code with casts, but this is good enough for now.

### DIFF
--- a/terps/magnetic/Generic/defs.h
+++ b/terps/magnetic/Generic/defs.h
@@ -38,15 +38,7 @@
 #include <stdint.h>
 
 typedef uint8_t type8;
-// The name implies a signed type, but this appears to generally be used
-// where "char" normally would (i.e. for C strings); that is, the fact
-// that it is signed does not appear to matter, and making it signed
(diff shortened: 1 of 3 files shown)


cspiegel:
@cibersheep This should fix the issue you were seeing on ARM.

cibersheep:
Indeed, it works. Thanks a lot!

PR #892, opened 2024-09-18
Use application name for shared MIME-info file

autumnontape:
[According to the freedesktop.org shared MIME-info spec:](https://specifications.freedesktop.org/shared-mime-info-spec/latest/ar01s02.html#s2_layout)

> Each application that wishes to contribute to the MIME database will install a single XML file, named after the application, into one of the three `<MIME>/packages/` directories (depending on where the user requested the application be installed).

This is extremely minor but makes Gargoyle a slightly cleaner package to have installed on one's system.


cspiegel:
Can't argue with that; thanks!

PR #899, opened 2024-10-10
Frankendrift updates

awlck:
... in which I scramble to add previously-overlooked common features (like transcripts), to the FrankenDrift Glk port in time for the planned Gargoyle release.


CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ if(UNIX AND NOT DIST_INSTALL)
     endif()
 endif()
 
-option(WITH_FRANKENDRIFT "Indicates that the FrankenDrift interpreter for ADRIFT 5 games will be made available" OFF)
+option(WITH_FRANKENDRIFT "Buld the FrankenDrift interpreter for ADRIFT 5 games (requires the .NET 8 SDK)" OFF)
```
cspiegel, line 73: Typo: "Buld" → "Build"

garglk/CMakeLists.txt
```diff
@@ -161,6 +161,9 @@ if(WITH_LAUNCHER)
     warnings(gargoyle)
 
     target_compile_definitions(gargoyle PRIVATE "GARGLKINI=\"${GARGLKINI}\"" "GARGLKPRE=\"${GARGLKPRE}\"")
+	if(WITH_FRANKENDRIFT)
+        target_compile_definitions(gargoyle PRIVATE WITH_FRANKENDRIFT=1)
+	endif()
```
cspiegel, line 166: This should be indented with spaces, not tabs.

cspiegel:
Apparently vcpkg just removed the "sndfile" package in favor of "libsndfile", so the Clang build doesn't work. I've fixed this in master; can you please rebase onto/merge master so the build can run here?

cspiegel:
Looks good to me. This is still in draft mode, so whenever you're ready for merge, you can move it out of draft and I'll merge.

awlck:
Alright, that's about all from my side.

The one thing that remains non-functional for some reason is sound. My `glk_schannel_play_ext` calls return true, but no sound gets played. (Curiously, WinGlk returns false for these calls, so perhaps my code is doing something wrong but garglk is falsely returning success? Not sure.)

cspiegel:
> The one thing that remains non-functional for some reason is sound. My `glk_schannel_play_ext` calls return true, but no sound gets played. (Curiously, WinGlk returns false for these calls, so perhaps my code is doing something wrong but garglk is falsely returning success? Not sure.)

Gargoyle should only return true if repeats is 0, or when it successfully starts the audio; any errors before that should result in 0 being returned.  I assume sounds work in other interpreters built with Clang for MSVC?

One thing to do would be to build a minimal Glk program using your bindings, and see if you get sound.  For me, on Linux/Qt6, the following C++ program produces sound:

```c++
    auto chan = glk_schannel_create(0);
    glk_schannel_play_ext(chan, 15, 1, 0);
    while (1) {
        event_t event;
        glk_select(&event);
    }
```
You just need a supported audio file (e.g. AIFF) named SND15 in the current directory.  That may help narrow down where the problem is.

awlck:
> I assume sounds work in other interpreters built with Clang for MSVC?

Well, *duh*. I should have checked that, it does not.

I guess this can be merged and I'll be off figuring out where sound is going wrong, in general.

PR #900, opened 2024-10-11
Add an ENABLE_GPL3 knob, disabled by default

cspiegel:
xBRZ is GPLv3 only, but some interpreters are GPLv2 only. garglk is either 2 or 3, so building with xBRZ support forces it to be 3, and thus incompatible with GPLv2 interpreters.


cspiegel:
Closed in favor of #901.

PR #901, opened 2024-10-12
Build both GPLv2 and GPLv3 versions of libgarglk

cspiegel:
The xBRZ scaler is GPLv3 only, while some interpreters are GPLv2 only. If Gargoyle is built with xBRZ support, then GPLv2 interpreters linked against it can't be distributed.

Work around this by creating two versions of libgarglk. One (libgarglk, as in previous versions) contains the extra GPLv3 code. The other (libgarglk-gpl2) does not. Other than this, the two libraries are identical.


cspiegel:
This is an alternative to #900 that is a complete fix, i.e. doesn't require any intervention from packagers.

@awlck There's one problem I'm having with the Clang build, though. The following occurs:

```
FAILED: terps/alan2.exe 
C:\Windows\system32\cmd.exe /C "cd . && "C:\Program Files\CMake\bin\cmake.exe" -E vs_link_exe --intdir=terps\CMakeFiles\alan2.dir --rc=C:\PROGRA~2\WI3CF2~1\10\bin\100226~1.0\x64\rc.exe --mt=C:\PROGRA~2\WI3CF2~1\10\bin\100226~1.0\x64\mt.exe --manifests D:\a\garglk\garglk\terps\garglk.manifest -- C:\PROGRA~1\LLVM\bin\lld-link.exe /nologo terps\CMakeFiles\alan2.dir\alan2\arun.c.obj terps\CMakeFiles\alan2.dir\alan2\main.c.obj terps\CMakeFiles\alan2.dir\alan2\debug.c.obj terps\CMakeFiles\alan2.dir\alan2\args.c.obj terps\CMakeFiles\alan2.dir\alan2\exe.c.obj terps\CMakeFiles\alan2.dir\alan2\inter.c.obj terps\CMakeFiles\alan2.dir\alan2\parse.c.obj terps\CMakeFiles\alan2.dir\alan2\rules.c.obj terps\CMakeFiles\alan2.dir\alan2\stack.c.obj terps\CMakeFiles\alan2.dir\alan2\decode.c.obj terps\CMakeFiles\alan2.dir\alan2\term.c.obj terps\CMakeFiles\alan2.dir\alan2\reverse.c.obj terps\CMakeFiles\alan2.dir\alan2\readline.c.obj terps\CMakeFiles\alan2.dir\alan2\params.c.obj terps\CMakeFiles\alan2.dir\alan2\sysdep.c.obj terps\CMa
LINK Pass 1: command "C:\PROGRA~1\LLVM\bin\lld-link.exe /nologo terps\CMakeFiles\alan2.dir\alan2\arun.c.obj terps\CMakeFiles\alan2.dir\alan2\main.c.obj terps\CMakeFiles\alan2.dir\alan2\debug.c.obj terps\CMakeFiles\alan2.dir\alan2\args.c.obj terps\CMakeFiles\alan2.dir\alan2\exe.c.obj terps\CMakeFiles\alan2.dir\alan2\inter.c.obj terps\CMakeFiles\alan2.dir\alan2\parse.c.obj terps\CMakeFiles\alan2.dir\alan2\rules.c.obj terps\CMakeFiles\alan2.dir\alan2\stack.c.obj terps\CMakeFiles\alan2.dir\alan2\decode.c.obj terps\CMakeFiles\alan2.dir\alan2\term.c.obj terps\CMakeFiles\alan2.dir\alan2\reverse.c.obj terps\CMakeFiles\alan2.dir\alan2\readline.c.obj terps\CMakeFiles\alan2.dir\alan2\params.c.obj terps\CMakeFiles\alan2.dir\alan2\sysdep.c.obj terps\CMakeFiles\alan2.dir\alan2\glkstart.c.obj terps\CMakeFiles\alan2.dir\alan2\glkio.c.obj terps\CMakeFiles\alan2.dir\alan2\alan.version.c.obj terps\CMakeFiles\alan2.dir\__\garglk\icons.rc.res /out:terps\alan2.exe /implib:terps\alan2.lib /pdb:terps\alan2.pdb /version:0.0 /machine:
lld-link: warning: found both wWinMain and WinMain; using latter
lld-link: error: undefined symbol: main
>>> referenced by C:\Users\qt\work\qt\qtbase\src\entrypoint\qtentrypoint_win.cpp:45
>>>               Qt6EntryPoint.lib(qtentrypoint_win.cpp.obj):(int __cdecl qtEntryPoint(void))
>>> referenced by C:\Users\qt\work\qt\qtbase\src\entrypoint\qtentrypoint_win.cpp:54
>>>               Qt6EntryPoint.lib(qtentrypoint_win.cpp.obj):(int __cdecl qtEntryPoint(void))
```
If you have any time, would you be willing to take a look at it?

The main change here is that the garglk CMake target is now an object library called garglk-common, instead of a normal library, and two "real" libraries are created, both of which link garglk-common.

I'm not well-versed in Windows, but it seems like it's building this as a console app?  It wants there to be a "main", and it's linking in qtentrypoint, which providees WinMain and wWinMain, and calls main directly.  So somehow it's apparently getting the wrong idea about what kind of application this is now?  But that's the best I can figure out so far.

cspiegel:
@awlck Hold that thought... I fixed the Clang build, but in the process broke the Mac build.  So I'm just gonna be doing whack-a-mole for a bit here.

awlck:
Looks like `Qt6EntryPoint.lib` does essentially the same as the `wWinMain` function I wrote here: https://github.com/garglk/garglk/blob/551ff21dac8453576090ba666a904d24f48aa649/garglk/main.cpp#L40, so the two conflict.

Honestly, I'd be in favor of removing my version and just relying on the Qt lib if we can get that to be linked in reliably.

cspiegel:
> Honestly, I'd be in favor of removing my version and just relying on the Qt lib if we can get that to be linked in reliably.

That makes sense.  I've got the whole thing working now by expressly _disabling_ Qt's entry point for the interpreters. It seems that when linking was done "properly", Qt was already aware, somehow, not to add its own entry point. Must be some property that propagated... but then when splitting things off, that no longer propagated, and the Qt entry point was linked, resulting in this problem.

Now that I more or less understand what's happening, I'll see about using Qt's entry point. Qt5 does this differently, but if we just require Qt6 on MSVC (which, as I recall, may already be a requirement?) then it hardly matters.

cspiegel:
OK, with Qt's entry point used instead of Gargoyle's, we're in good shape.  I'll leave this open for a bit so anybody interested can take a look, but unless there's major problems, I'll get this in and close #900.

awlck:
> if we just require Qt6 on MSVC (which, as I recall, may already be a requirement?)

I don't think Qt6 a requirement per se, it's just that I never bothered to test with Qt5 since it's no longer supported with first-party binaries. (A few years back, Qt changed their support policy to only provide official LTS builds to paying customers. So the newest version of Qt5 I can get through their installer is 5.15.2.)

cspiegel:
> I don't think Qt6 a requirement per se, it's just that I never bothered to test with Qt5 since it's no longer supported with first-party binaries. (A few years back, Qt changed their support policy to only provide official LTS builds to paying customers. So the newest version of Qt5 I can get through their installer is 5.15.2.)

I'm fine with Qt5 being unsupported/untested in this case. If it works, great, if not, Qt5 is basically obsolete at this point anyway, as you note.

cspiegel:
@autumnontape Is this sufficient for your needs?  This should require absolutely no work on your part to produce a legally-distributable set of binaries, as long as you use shared libraries.

autumnontape:
@cspiegel I tried building packages of this PR for Arch and Chimera Linux. Both seem to work as before (including interpreters that link to both versions of the library), and it did indeed require no work on my part on either. This is a wonderful solution from a packaging perspective! I really appreciate all the effort you go to for packagers.

I notice the tadsr executable links to the GPLv3 `libgarglk.so`. Can tads-runner be distributed as GPLv3, or is that an oversight?

cspiegel:
> I notice the tadsr executable links to the GPLv3 `libgarglk.so`. Can tads-runner be distributed as GPLv3, or is that an oversight?

That was an oversight, thanks!

PR #902, opened 2024-10-20
Fix Qt6 sound playback on Windows

awlck:
This fixes sounds not playing on Windows when using Qt6 as the playback backend. (This was the underlying issue of my FrankenDrift sound implementation not working, as discussed in #899.)

(MacOS seems to suffer from a similar but distinct bug that I haven't been able to track down yet, but as I understand, SDL is the default playback backend on the Mac so it's not critical right now. I can open an issue for that if desired.)

(Initially keeping this as a draft until I get a chance to verify that I haven't broken anything on Linux in the process.)

### DIFF
--- a/garglk/sndqt.cpp
+++ b/garglk/sndqt.cpp
@@ -236,6 +236,20 @@ public:
         m_buffer_size = size;
     }
 
+#ifdef HAS_QT6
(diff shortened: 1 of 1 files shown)


cspiegel:
This works just fine on Linux. Thanks!

awlck:
Thanks for the confirmation!

PR #920, opened 2025-03-29
Set CMAKE_INSTALL_RPATH to handle library paths correctly

shmup:
This resolves: error while loading shared libraries: libgarglk.so: cannot open shared object file: No such file or directory

Now, is this the right approach? I'm unsure!

In my case, this allowed it to find the libs in `/usr/lib64`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,9 @@ string(STRIP "${GARGOYLE_VERSION}" GARGOYLE_VERSION)
 # CI builds which include the current Git revision as the version.
 project(garglk)
 
+include(GNUInstallDirs)
+set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
+
 if(POLICY CMP0069)
     cmake_policy(SET CMP0069 NEW)
     include(CheckIPOSupported)


cspiegel:
Ideally the rpath shouldn't need to be set. If CMake is installing to `/usr/lib64` then your system should be looking in `/usr/lib64` for shared libraries, e.g. using `/etc/ld.so.conf` on glibc-based systems. Do you have any other libraries in `/usr/lib64`? Are they found by the runtime linker? It'd be kind of weird if `libgarglk.so` is the only library _not_ being found there.

That said, if there _are_ systems out there which have CMake installing to non-standard locations, it would be nice to work with them. So setting the rpath is a reasonable solution there, but I think I'd want it gated behind a build flag: rpath is kind of a sledgehammer and prevents users from doing things like `LD_LIBRARY_PATH`.

So I'm not against this per se, though I would be interested in knowing about how `/usr/lib64` looks on your system, i.e. if there are any other libraries there that are found.

If we do go forward with setting rpath, though, only `CMAKE_INSTALL_RPATH` should be set. `CMAKE_BUILD_WITH_INSTALL_RPATH` will set the rpath to `/usr/lib` (or whatever) even when the programs are in the build directory, which will break testing: when running out of the build directory, interpreters wouldn't be able to find an updated `libgarglk.so`, instead looking at the system version. For most people this doesn't matter, but it'd make development a lot more difficult.

And I don't think `CMAKE_INSTALL_RPATH_USE_LINK_PATH` is needed either. From what I can tell that will slurp in all library directories CMake can find and add them to the rpath. But it should be enough to set the install path to where `libgarglk.so` is. If other system libraries aren't found by the runtime linker, that's a wholly broken system.

shmup:
hey @cspiegel thanks for the reply. I see that on my Fedora system, libgarglk.so is installed to `/usr/local/lib64/` which isn't in my system's dynamic linker search paths. also, in general. many rpm-based distros use lib64 (and lib for 32bit)

this isn't really a territory I'm savvy in, would defer to you for all thoughts haha

though you're right, I only needed:

```
include(GNUInstallDirs)
set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
```

I did need the include, fwiw. again, out of my realm here :)

cspiegel:
> I see that on my Fedora system, libgarglk.so is installed to /usr/local/lib64/ which isn't in my system's dynamic linker search paths.

CMake, like autotools, defaults to `/usr/local` so as not to interfere with the system which is in `/usr`. But! As you've noticed, that's often not part of the runtime search path. This is a fairly common scenario.

I think the appropriate approach here would be one of two things:

1. Add `/usr/local/lib64` to your search path. If Fedora supports `/etc/ld.so.conf.d` then ideally you'd just create a new file there and add `/usr/local/lib64` to it.
2. Use the included `gargoyle-buildrpm.sh` script to create an RPM and install it. I think this is the better solution since the system will track the package and allow for easy upgrades as you build newer versions.

If one of those works, then that's better than unconditionally adding rpath support, though I'd be OK with adding a conditional for it; I'd leave that up to you. i.e. if neither solution works, we can add an `option()` that will set the rpath. If one of the solutions _does_ work we can leave it, or, if you'd like, we can still add the conditional rpath support.

cspiegel:
Closing because I don't think an rpath setting is required. Feel free to reopen/create a new PR if you think otherwise.

PR #931, opened 2025-05-08
Sync level9

cspiegel:


cspiegel:
@balyuf After a small hiccup, I've got your Level9 changes merged here. Thanks!

balyuf:
Unfortunately, through comparing my entries with the ones in the Babel project, I discovered that I'd accidentally introduced a duplicate entry.  So, a new PR upstream was accepted for a one-liner: https://github.com/DavidKinder/Level9/pull/29 ... What should we do here?  Quickly fix this on the master, go through the whole process again, or leave it like that until there is more to merge later?  This is a minor fix after all, as it is more to keep things clean than to fix an issue (the detection is fine) ...

cspiegel:
Since it has no effect on gameplay/detection, there's no harm in just keeping the duplicate here. Eventually when 5.2 is released we'll pull that in and all will be good.

balyuf:
> Since it has no effect on gameplay/detection, there's no harm in just keeping the duplicate here. Eventually when 5.2 is released we'll pull that in and all will be good.

Well, I am not sure if you're automatically informed, but David has released 5.2 of the Level 9 interpreter ...
See https://github.com/DavidKinder/Level9/releases/tag/5.2

PR #948, opened 2025-08-16
Persist flow breaks across window reflows

### LINKED ISSUE #222: glk_window_flow_break is handled incorrectly
Per [spec](http://www.eblong.com/zarf/glk/glk-spec-074_7.html):

>  When a text buffer window is resized, a flow-break behaves cleverly; it may become active or inactive as necessary. You can consider this function to insert an invisible mark in the text stream. The mark works out how many newlines it needs to be whenever the text is formatted for display.

Gargoyle appears to insert hard breaks.  If the window is resized, the number of breaks doesn't change, leading to text errantly wrapping around the image or excessive whitespace.

salty-horse:
This is an attempt to fix #222.

Gargoyle currently stores the "text stream" as "lines" object, and those lines are determined using the current window dimensions and font size, etc. Additionally, it doesn't support inline images, and allows for only one inline image per such "line".

When the window is resized, and it needs to reflow the text, it extract the data from those old "lines" and creates new ones.

The current implementation of `flow_break` emulates it by sticking a bunch of newlines as raw text. When the display is reflowed, they're still treated as regular newlines.

I tried changing that to a "flow break" marking on the line, which says that at position X in the text, insert a flow break. It relies on the fact that when inline images are inserted, Gargoyle currently sticks them in their own separate line. So this flow break is always assumed to be called AFTER the inline images.

I tested it with the sample program linked in the issue - note that it inserts a flow-break before the margin image, which is ignored as it's reset in `put_picture`.

I also tested with [Everybody Dies](https://ifdb.org/viewgame?id=lyblvftb8xtlo0a1), which opens with an inline image followed by a flow break.

I also tested it with [imagetest.gblorb](https://eblong.com/zarf/glulx/imagetest.gblorb), which displays the image like it did before the change.

Tests include printing the images with various initial window sizes, and then trying to resize the window to make sure it still flows correctly.

I did NOT test flow-break commands between two text strings.

Please let me know if it breaks in other situations where Gargoyle used to work in the past.

### DIFF
--- a/garglk/garglk.h
+++ b/garglk/garglk.h
@@ -933,6 +933,7 @@ struct tbline_t {
         chars.fill(' ');
     }
     int len = 0;
+    int flow_break_pos = -1;
     bool newline = false, dirty = false, repaint = false;
     std::shared_ptr<picture_t> lpic, rpic;
     glui32 lhyper = 0, rhyper = 0;
--- a/garglk/wintext.cpp
+++ b/garglk/wintext.cpp
@@ -37,6 +37,8 @@ static void
 put_text_uni(window_textbuffer_t *dwin, glui32 *buf, int len, int pos, int oldlen);
 static bool
 put_picture(window_textbuffer_t *dwin, const std::shared_ptr<picture_t> &pic, glui32 align, glui32 linkval);
+static void
+scrolloneline(window_textbuffer_t *dwin, bool forced, bool flow_break);
 
 static void touch(window_textbuffer_t *dwin, int line)
 {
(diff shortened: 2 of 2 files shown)


cspiegel:
> Here's a better representation of "inline" data in a line. It can be done instead of this PR, or as a next step:

I keep planning on doing a Gargoyle release "imminently" (that's now becoming more and more the wrong word), so I think it's reasonable to merge this to have a fix for the release, and if an improved representation for inline data can be done before the release, great; otherwise, at least there's a fix for #222.

Unless you have any compelling reason not to, I'll merge this in and if you'd like, you can create an issue laying out how you think `tbline_t` ought to be updated.

salty-horse:
I started laying out the data structures, but haven't updated the code yet.
It will be done in stages, anyway. Supporting inline images and variable-height lines would require more changes.

I'm OK with merging this if it's good enough.

PR #987, opened 2026-06-18
Fix blurry fractional scaling on Wayland

electronstudio:
Allows use of fractional display scaling (e.g. 1.25) on Wayland.  Only tested on QT6, not QT5.

It doesn't attempt to set the gli_backingscalefactor (see comment in code) so user will still need to set `zoom 1.25` in config file.

### DIFF
--- a/garglk/config.cpp
+++ b/garglk/config.cpp
@@ -85,6 +85,27 @@ std::string garglk::ConfigFile::format_type() const {
     }
 }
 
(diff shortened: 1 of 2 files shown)


cspiegel:
I made small changes; details, such as they are, are in commit messages, but in short:

* Include cmath for std::round(), and consistently use std::round()
* Always use devicePixelRatioF() to simplify the code

Thanks for the PR!

PR #994, opened 2026-07-16
Add support for Glk `stylehint_Justification`

dfabulich:


dfabulich:
This makes it possible to render status lines without a text grid. The status line can be a split window with three panes, one left aligned, one center aligned, and one right aligned. (I think it looks much nicer that way.)

image

cspiegel:
Looks great, thank you!

PR #996, opened 2026-07-21
Fix Wayland fractional-scale first-frame buffer sizing race

### LINKED ISSUE #995: Wayland: devicePixelRatioF() returns rounded scale, not true fractional value
## Summary

Testing the recent `eed8b7a3` / `8d242b8d` / `9e560518` commits (not yet in a tagged release) on a real fractional Wayland output scale (125%, KWin 6.7.3), text still renders soft/blurry rather than crisp. Root cause looks to be that `devicePixelRatioF()` on Wayland only ever returns the *rounded* integer buffer scale (2.0 for 125%), not the true fractional value, even though the compositor negotiation for the true value succeeds at the protocol level.

## Evidence

Running with `WAYLAND_DEBUG=1`, the client correctly binds `wp_fractional_scale_manager_v1` and receives the real value:

```
wp_fractional_scale_manager_v1#7.get_fractional_scale(new id wp_fractional_scale_v1#39, wl_surface#38)
wp_fractional_scale_v1#39.preferred_scale(150)   # 150/120 = 1.25, the real output scale
```

But `View::devicePixelRatioF()` inside the app itself only ever returns `2.0` for the same window, for the lifetime of the process. So `resizeEvent()`'s physical-pixel buffer sizing (`newwid * dpr`) ends up using 2.0 instead of the true 1.25, and the resulting buffer gets scaled by the compositor to fit the real output: exactly the blur the recent commits were meant to fix, just with a smaller error margin than before (2.0 vs. the old 1.0).

## Secondary issue: first-frame timing

Independently of the above, `preferred_scale` arrives *asynchronously*, generally after the window's first `resizeEvent()` has already fired and sized `gli_image_rgb`. Since nothing reacts to a devicePixelRatio-only change (no logical resize), and the buffer-size check in `resizeEvent()` is an equality short-circuit, a DPR change with no accompanying resize is silently dropped. In my testing this occasionally produced a fully blank window on first launch (buffer allocated at the wrong physical size, and nothing ever corrects it) until some unrelated event forced a second resize.

## What I found actually fixes it (locally)

Reading the true scale via QtWaylandClient's private `QWaylandWindow::scale()` instead of `devicePixelRatioF()` gets the correct 1.25 and produces genuinely crisp output, once combined with re-triggering the buffer-size calculation shortly after the window is shown (to pick up the true value once it's arrived). I'm not proposing this as a patch, since it pulls in `Qt6WaylandClientPrivate` (unstable ABI, tied to the exact Qt point release), which isn't something that makes sense for a project shipping across Windows/macOS/Linux. But it does confirm precisely where the gap is: Qt's own internal Wayland window object tracks the true fractional value (it's what it uses to size the real on-screen surface via `wp_viewport`), it's just not exposed through any public Qt API (`QScreen`, `QWindow::devicePixelRatio()`, etc.).

## Environment

- Gargoyle 2026.1.1 + the three commits above, built locally
- Qt 6.11.1, KWin 6.7.3, output scale 125%
- Confirmed via raw `WAYLAND_DEBUG=1` traces, not just visual inspection

Happy to share full debug logs, or test any patch attempts, if useful.

senjinthedragon:
Fixes the remaining first-frame issue from #995: on Wayland with a fractional output scale, `preferred_scale` arrives asynchronously and generally hasn't been processed yet when the window's first `resizeEvent()` fires, so `devicePixelRatioF()` still reports the fallback/rounded value at that point. Since later `resizeEvent()` calls at the same logical size are a no-op, the wrong physical buffer size from that first call sticks around even though `devicePixelRatioF()` itself corrects on every call after.

Two commits:
1. Extracts the buffer-sizing logic into `updateBufferSize()` and re-runs it once, shortly after the window is first shown, using a timer.
2. Per review feedback, replaces the timer with `QEvent::DevicePixelRatioChange` (Qt 6.6+) where available, since a fixed delay is inherently racy. Kept the timer as a fallback for Qt < 6.6.

Tested locally on Qt 6.11.1 (KWin 6.7.3, 125% output scale), confirmed crisp rendering on first launch with no leftover blur/blank window.

Fixes #995

### DIFF
--- a/garglk/sysqt.h
+++ b/garglk/sysqt.h
@@ -61,6 +61,9 @@ protected:
     void closeEvent(QCloseEvent *) override;
     void resizeEvent(QResizeEvent *) override;
     void moveEvent(QMoveEvent *) override;
+#if QT_VERSION >= QT_VERSION_CHECK(6, 6, 0)
+    bool event(QEvent *) override;
+#endif
 
 private:
     // Resizes gli_image_rgb to the physical-pixel size for the given
--- a/garglk/sysqt.h
+++ b/garglk/sysqt.h
@@ -9,6 +9,7 @@
 #include <QPaintEvent>
 #include <QResizeEvent>
 #include <QSettings>
+#include <QShowEvent>
 #include <QTimer>
 #include <QWheelEvent>
 #include <QWidget>
@@ -56,11 +57,18 @@ public:
     const QSettings *settings() { return m_settings; }
 
 protected:
+    void showEvent(QShowEvent *) override;
     void closeEvent(QCloseEvent *) override;
     void resizeEvent(QResizeEvent *) override;
     void moveEvent(QMoveEvent *) override;
 
 private:
+    // Resizes gli_image_rgb to the physical-pixel size for the given
+    // logical size at the view's *current* effective_dpr(). Shared by
+    // resizeEvent() and the delayed call in showEvent() - see the
+    // comment there for why a second call is needed.
+    void updateBufferSize(const QSize &logicalSize);
+
     View *const m_view;
     QTimer *const m_timer;
     QSettings *const m_settings;
--- a/garglk/sysqt.h
+++ b/garglk/sysqt.h
@@ -67,9 +67,9 @@ protected:
 
 private:
     // Resizes gli_image_rgb to the physical-pixel size for the given
-    // logical size at the view's *current* effective_dpr(). Shared by
-    // resizeEvent() and the delayed call in showEvent() - see the
-    // comment there for why a second call is needed.
+    // logical size at the view's current devicePixelRatioF(). Shared by
+    // resizeEvent() and the extra call triggered from showEvent()/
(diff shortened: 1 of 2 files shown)


garglk/sysqt.h
```diff
@@ -56,11 +57,21 @@ class Window : public QMainWindow {
...
     void closeEvent(QCloseEvent *) override;
     void resizeEvent(QResizeEvent *) override;
     void moveEvent(QMoveEvent *) override;
+#if QT_VERSION >= QT_VERSION_CHECK(6, 6, 0)
+    bool event(QEvent *) override;
+#endif
 
 private:
+    // Resizes gli_image_rgb to the physical-pixel size for the given
+    // logical size at the view's *current* effective_dpr(). Shared by
```
cspiegel: It's minor since it's just a comment, but there's no `effective_dpr()` function in this code.

senjinthedragon: Good catch, fixed. Leftover from the earlier private-API version of this comment.

PR #997, opened 2026-07-27
Add optional gamepad/Steam Controller input support

alytaphoenix:
Adds `WITH_CONTROLLER` (CMake option, off by default), which lets Gargoyle read input from a connected game controller via SDL2 — D-pad → arrows, A → Return, B → Escape, shoulders → Page Up/Down — usable standalone or as a non-Steam game with Steam Input. Runtime-togglable via a new `controller` config option, independent of the `SOUND` backend choice.

Only wired up for the Qt interface (Windows and Linux always use it; on macOS it's opt-in and the Cocoa build stays unaffected — documented in INSTALL.md). Not currently supported together with `SOUND=SDL3` (SDL2/SDL3 can't safely link into the same process; documented and enforced at configure time).

**Testing:** Built and reviewed across every `SOUND`/`WITH_CONTROLLER` combination, and manually verified end-to-end via a Qt6 macOS build with a real controller (button dispatch, config toggle, focus gating). **Not yet verified on the actual target platform (Linux/SteamOS)** or against Steam Input's translation layer specifically — macOS testing hit environment-specific dead ends unrelated to this code (details on request). Would appreciate testing from anyone with a Deck or Linux+controller setup before merge.

Built with AI-assisted tooling (Claude Code); human-reviewed throughout, happy to adjust process if that's a concern.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -118,6 +118,22 @@ In addition, Gargoyle supports the following options:
   for SDL3 sound support, "QT", for Qt sound support, and "NONE", for no sound
   support. "SDL" is accepted as a backward-compatible alias for "SDL2".
 
+- `WITH_CONTROLLER`: If true, enable gamepad/Steam Controller input support,
+  using SDL2. Defaults to false. Can be disabled at runtime, without a
+  rebuild, via the `controller` configuration file option.
+
+    - Not currently supported together with `SOUND=SDL3`, since SDL2 and
+      SDL3 cannot safely be linked into the same process; use `SOUND=SDL2`,
+      `QT`, or `NONE` instead, or leave this option off.
+
+    - Only takes effect with `INTERFACE=QT` (the only choice on Windows and
+      Linux, and optional on Apple). With `INTERFACE=COCOA`, the SDL
+      controller subsystem still initializes, but nothing polls it, so
+      gamepad input is silently inert rather than broken. Extending this to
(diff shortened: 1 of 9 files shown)


cspiegel:
Can I ask what the use-case is for a controller with Gargoyle? Apart from novelties like Freefall, you'll still need a keyboard to do anything useful.

alytaphoenix:
Fair question, and you're right that the parser itself still needs a keyboard — this isn't trying to replace typing.

The actual use case is Steam Deck / couch / TV setups: Big Picture mode, a controller in hand, on-screen keyboard for typing commands (already how Deck users interact with any text field). What's annoying in that setup isn't typing a command, it's everything *around* it — hitting Return to continue a "[MORE]" prompt, paging back through scrollback to re-read something, dismissing a dialog — all of which currently mean picking up a keyboard (or fighting the on-screen one) for a single keypress. That's the gap this closes: D-pad/A/B/shoulders cover exactly that navigation surface (movement, confirm, cancel, page up/down), the on-screen keyboard still handles actual parser input.

Rebased onto current master (one conflict in README.md, both additive — resolved cleanly) and pushed. Still stands as-is on testing: solid on a Qt6 macOS build with a real controller, but genuinely unverified on the actual target platform (Linux/SteamOS) or against Steam Input's translation layer specifically — if anyone here has a Deck or Linux+controller setup, that testing would be the thing that actually de-risks this before merge.

PR #1000, opened 2026-08-09
Add the SCARIER interpreter from Spatterlight

cspiegel:
This is not built by default. It needs `-DWITH_SCARIER=ON -DWITH_SCARE=OFF`, and cannot be built alongside FrankenDrift.

This is imported from Spatterlight 1.5.2.


dfabulich:
I filed a PR https://github.com/cspiegel/garglk/pull/2 targeting cspiegel:add-scarier, updating `gargoyle_osx.sh` to support passing `GARGOYLE_CMAKE_EXTRAS` so I can build Gargoyle for macOS with Scarier.

PR #1001, opened 2026-08-10
DRAFT: Update Scarier, implement Danni's CSS Glk extension

dfabulich:
1. Update Scarier from https://github.com/angstsmurf/spatterlight/commit/d14234ef
2. Copy rich-text-css support from https://github.com/angstsmurf/spatterlight/pull/135
3. Implement Dannii's CSS Basic Glk extension, based on https://github.com/angstsmurf/spatterlight/pull/135 (using arbitrarily selected gestalt/function IDs).
4. Implement arbitrary custom fonts. (In an earlier draft of this PR, it only did monospace vs. proportional).

Like https://github.com/angstsmurf/spatterlight/pull/135 the role of this PR is to act as a proof of concept for Dannii's Basic CSS Glk extension. I think this code looks pretty good.

image

### DIFF
--- a/garglk/garglk.h
+++ b/garglk/garglk.h
@@ -563,19 +563,37 @@ struct picture_t {
     bool scaled;
 };
 
+// A set of CSS declarations, as used by the CSS Basic extension. Keys
+// are property names, lowercased; values are the property values as
+// provided by the game, with surrounding whitespace stripped.
+using CssProps = std::map<std::string, std::string>;
+
 struct style_t {
     FontFace font;
     Color bg;
     Color fg;
     bool reverse;
     glui32 justification = stylehint_just_LeftFlush;
+    std::optional<double> size;
+    bool underline = false;
+    double margin_left = 0;
+    double margin_right = 0;
+    double text_indent = 0;
+    // CSS_Paragraph background-color: content-box fill (not glyph-run bg).
+    std::optional<Color> para_bg;
 
     bool operator==(const style_t &other) const {
         return font == other.font &&
                bg == other.bg &&
                fg == other.fg &&
                reverse == other.reverse &&
-               justification == other.justification;
+               justification == other.justification &&
+               size == other.size &&
+               underline == other.underline &&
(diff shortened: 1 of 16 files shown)


cspiegel:
Once #1000 is in, I will push out the `upstream-scarier` branch, which is the method used to bring in upstream changes. Ideally a new Spatterlight will be tagged in not too long so that I'm not pulling in arbitrary Git hashes, but if need be, I can.

I've got a couple comments on some of the proposed changes from a cursory glance, and I'll do a more in-depth review when I have a bit more time.

garglk/garglk.h
```diff
@@ -806,22 +821,55 @@ struct glk_fileref_struct {
...
+
+    // CSS Basic overrides, packed for the per-character attr buffer.
+    // Tri-states use -1 = unset, 0 = false, 1 = true. size == 0 means
+    // unset (use the style / config default). justification == 0xff
+    // means unset.
+    std::int8_t bold = -1;
+    std::int8_t italic = -1;
+    std::int8_t monospace = -1;
+    std::uint8_t justification = 0xff;
+    float size = 0;
```
cspiegel: These should all be std::optional (and the bold/italic/monospace std::optional<bool>). Those plus justification are mentioned in the comment (specifically calling out "unset" indicates we want optional); and from what I can see "size" being 0 means unset as well.

garglk/garglk.h
```diff
@@ -806,22 +821,55 @@ struct glk_fileref_struct {
 
 struct attr_t {
     bool reverse = false;
+    bool underline = false;
```
cspiegel: Is there a reason this is a plain bool vs the tristates/optional for the other CSS overrides?

dfabulich:
Resolved feedback, updated Scarier, and implemented arbitrary custom fonts.

(We still can't really merge this until Dannii chooses a gestalt ID, I think…)

dfabulich:
Replaced with #1035 and #1036

PR #1002, opened 2026-08-10
Scarier update with text colours

angstsmurf:
Just opening this as a draft to let you know that I've (well, my AI has) added Adrift text colours to Scarier, using the zcolor system, and tested them in Gargoyle, where they seem to work fine. They need the -c command line switch or the GLK COLOUR command.


cspiegel:
Is there an imminent Spatterlight 1.5.3 release, or is that indeterminate? I ask because I _prefer_ to upstream releases instead of arbitrary commits; but if it's some distant unknown point, then the Git commit's not a huge deal, especially since Scarier is in a decent amount of flux.

Updating upstream is kind of finnicky here in Gargoyle. It'll have to be done in the `upstream-scarier` branch. I keep all upstream interpreters in their own, orphaned branches, so that it's trivial to see which changes Gargoyle has laid on top, and so it's much easier _not_ to lose Gargoyle-specific changes during an upstream merge. It does mean the process is more involved though.

If you indicate that taking Spatterlight's current HEAD is the best approach, I can do so via the upstream mechanism.

angstsmurf:
I'd say the next Spatterlight release is not imminent. There are some things I'd like to polish first, such as the Alan 2 interpreter.

This pull request was mostly just to let you know that there has been major Scarier work done since the version you merged, including stuff that is specifically meant to make it easier to incorporate into Gargoyle.

Now you know, so I suppose this can be closed.

PR #1004, opened 2026-08-11
Basic menu bar (… and an IPC system?!?)

### LINKED ISSUE #1003: Add a menu bar
On Windows/Linux, Gargoyle currently has no buttons/menus once the game starts running. The window has a title bar, with standard buttons to minimize, maximize, and close the window, but it has no menu bar, and I think this makes it hard to discover how to use Gargoyle.

On macOS, Gargoyle _does_ have a basic menu bar, and I think it really helps:

* There's a "Gargoyle --> Settings…" menu item that opens ~/garglk.ini. As far as I know, there's no user-visible way to access the settings file (or to even know it exists) on Windows/Linux.
* A "File --> Open…" command lets you select a game, and a "File --> Open Recent >" menu lets you select games you've recently opened. On Windows/Linux, Gargoyle pops up an Open dialog on start, but there's no user-visible way to open another game except to quit Gargoyle.
* There's a "Help --> Gargoyle Help" menu, but it doesn't do anything except pop up a dialog that says, "Help isn’t available for Gargoyle." Presumably if there were any online documentation for Gargoyle #637 the menu could open an URL in the browser.

I think a very basic File menu would be useful on Windows/Linux. But, there's more that could go in there!

* The File menu could have options to Save, Restore, and Restart
* A File --> Export Transcript option
* The Edit menu could have an option to Find text in the transcript
* A View menu could allow users to configure some basic settings in UI
    * Text size / Zoom
    * Enable/disable graphics and sound
    * Override game colors
* If we ever implement my proposed [Automap Glk extension ](https://intfiction.org/t/automaps-in-a-glk-extension/81019), there could be a View --> Show Map menu.

WDYT? Would you consider a PR with a menu bar in it for Windows/Linux?

### LINKED ISSUE #1006: Launch Windows/Linux interpreters as IPC subprocesses
* Apropos #1003 
* Apropos #1004 

The number 1 most obvious thing you'd want to put in a basic menu bar is the option to open a game, but this is surprisingly thorny, because the interpreters are all `.exe` executables on Windows.

When you launch `gargoyle.exe`, it prompts you to select a game file, chooses a terp, then launches a terp process, e.g. `glulxe.exe` or `bocfel.exe`; the terps themselves show the game window.

If you then try to open another game in #1004 6fe633202dc8e99f717fff29429948455e3336da, it has to launch a whole other copy of Gargoyle. As a result, the Windows dock doesn't perceive those windows as belonging to the "same app."

On macOS, which already has a menu bar, there's an IPC subprocess system for launching interpreters as subprocess that communicate with the parent process.

That IPC system should be ported to Windows and Linux so we can have a reasonable File --> Open experience.

dfabulich:
* Fixes #1003
* Fixes #1006 

It's kinda wild to fix #1006 and #1003 in the same PR, but that's kinda what's needed to implement a basic "Open" menu.

There are two new `garglk.ini` boolean settings, `menu_bar` and `ipc`. `menu_bar` enables/disables the menu bar UI; `ipc` enables/disables the IPC system.

If you approve this PR, hopefully other menu items will be possible without such major refactorings.

EDIT: I did all of my manual testing for this branch on QT 6, on Windows 11 aarch64 and Ubuntu 24 LTS ARM, in VMs on an Apple Silicon Mac.

image

### DIFF
--- a/garglk/CMakeLists.txt
+++ b/garglk/CMakeLists.txt
@@ -99,7 +99,7 @@ if(INTERFACE STREQUAL "QT")
     set_property(SOURCE ${GARGLKINI_CXX} ${THEME_DARK_CXX} ${THEME_DARK_H} ${THEME_LIGHT_CXX} ${THEME_LIGHT_H} PROPERTY SKIP_AUTOMOC ON)
 
     if(NOT WITH_NATIVE_FILE_DIALOGS)
-        set_property(SOURCE sysqt.cpp launchqt.cpp APPEND PROPERTY COMPILE_DEFINITIONS GARGLK_CONFIG_NO_NATIVE_FILE_DIALOGS)
+        set_property(SOURCE sysqt.cpp launchqt.cpp menubarqt.cpp APPEND PROPERTY COMPILE_DEFINITIONS GARGLK_CONFIG_NO_NATIVE_FILE_DIALOGS)
     endif()
 endif()
 
@@ -149,7 +149,7 @@ if(INTERFACE STREQUAL "COCOA")
     target_sources(garglk-common PRIVATE sysmac.mm)
     target_compile_options(garglk-common PRIVATE "-Wno-deprecated-declarations")
 else()
-    target_sources(garglk-common PRIVATE sysqt.cpp)
+    target_sources(garglk-common PRIVATE sysqt.cpp ipc.cpp ipclientqt.cpp menubarqt.cpp)
     find_package(Threads REQUIRED)
     target_link_libraries(garglk-common PRIVATE ${CMAKE_THREAD_LIBS_INIT})
 endif()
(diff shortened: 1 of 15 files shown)


dfabulich:
Rebased onto `master`, tested in QT6 for Windows and Linux.

cspiegel:
I'm not ignoring this! I just have two local branches I've been working on, on and off for the last severael months, that do very similar things to this. The big difference is that my IPC system was Mac-only, and implemented largely as the first step on a path to retiring the Cocoa interface. But it's Mac-only in a very loose sense. Since Qt is so portable, it should be pretty easy to reconcile your Windows/Linux-based code (the way we're both doing IPC is quite similar).

I'm working now on integrating your optional/selectable IPC with my code to provide a system that's supported across all operating systems. If this lands, I think Cocoa will only have one more release where it survives. Not that there's anything inherently wrong with the code, but the idea of maintaining one single piece of UI code is very attractive.

dfabulich:
I rebased this PR again. I believe the Clang failure is flakiness and not an actual failure.

Apropos QT + macOS, I filed a stacked PR #1034 based on this PR. If your goal is to drop Cocoa, I think it would make sense to merge this PR and then #1034.

But, I gotta say, the macOS app #1034 is not as good as the Cocoa version right now, because it's missing a lot of functionality. #1034 could merge today (just having a QT build option won't hurt anything), but it would be a lot of work to make the QT version feel right (like a "Mac-assed Mac app"), or even _as_ right as the Cocoa version we already have.

dfabulich:
Added a few more menu items:
* File -> Close (in addition to File -> Exit)
* Edit Cut/Copy/Paste
* Help -> About Gargoyle

PR #1008, opened 2026-08-13
Improve Windows build

dfabulich:
When building on Windows, copy executables and DLLs directly into cmake-build, so you can double-click on `gargoyle.exe` to test.

### DIFF
--- a/.github/workflows/windows-clang.yml
+++ b/.github/workflows/windows-clang.yml
@@ -69,12 +69,11 @@ jobs:
       - name: Assemble results
         run: |
           mkdir gargoyle-staging
-          cp cmake-build\garglk\gargoyle.exe gargoyle-staging
-          cp cmake-build\garglk\*.dll gargoyle-staging
+          # Native Windows builds colocate runtimes in the CMake build directory root.
+          cp cmake-build\*.exe gargoyle-staging
+          cp cmake-build\*.dll gargoyle-staging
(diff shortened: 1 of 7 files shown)


cspiegel:
I've taken this plan but implemented it slightly differently, in #1015. The basic plan is the same: install DLLs and EXEs to the same build dir to make life easier. But the differences:

* The major one: `windeployqt` is _not_ run as a post-build step. That would do largely duplicate, slower work each and every time you rebuild. Instead, there's now a "deployqt" target which you can run once (until you blow away the build dir) and have it copy everything. Then future runs will keep working.
* Looking up `windeployqt` is done the way Qt does it, so should be more bulletproof.
* `CMAKE_RUNTIME_OUTPUT_DIRECTORY_*` is not touched, so that debug/release don't overwrite each other.
* `CMAKE_RUNTIME_OUTPUT_DIRECTORY` is conditionally set: the user can override it on the command-line.

I also didn't carry over the plugin exclusions from `winbuildrelease.ps1` (if those need to change it should be a separate PR), and no `.gitignore` updates, since `/cmake-build` is not a "standard" location here (it's used in the Clang CI but that doesn't affect Git here).

If there are issues with it vs your expectations, let me know and we can get it in line with how you're thinking.

dfabulich:
Let's merge #1015 instead.

PR #1030, opened 2026-09-04
Add a Dockerfile for cross-compiling MinGW

dfabulich:
With this, you can `./gargoyle_docker_mingw.sh` with nothing pre-installed but Docker itself.


cspiegel:
Very cool! Ultimately I think some organization of all the build scripts/files is in order (lots of stuff just heaped in the root), but no harm with the status quo for the time being.

PR #1035, opened 2026-09-13
Implement Dannii's CSS Basic Glk extension

dfabulich:
Reimplemented based on PR https://github.com/angstsmurf/spatterlight/pull/135

https://curiousdannii.github.io/if/glk-proposals.html#css

Test with https://github.com/dfabulich/glk-css-test and with #1036 

There's also a separate commit to add support for `font-family`; it's quite a bit of code in its own right.

Screenshot 2026-09-13 at 2 10 12 AM

### DIFF
--- a/garglk/CMakeLists.txt
+++ b/garglk/CMakeLists.txt
@@ -109,7 +109,7 @@ else()
     set(IMGLOAD imgload-system.cpp)
 endif()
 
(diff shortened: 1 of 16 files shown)


dfabulich:
The API is churning https://intfiction.org/t/glk-extension-proposal-css/66228

